PR #58, opened 2026-04-04
feat: add server_version to all JSON tool outputs

kumaakh:
Instead of adding a dedicated `server_info` tool, inject `server_version` into every JSON-format tool response. This reduces tool count while making the running version visible whenever structured output is requested.

**Changes:**
- `list_members(format="json")` — adds `server_version` field
- `member_detail(format="json")` — adds `server_version` field (included even on early-exit offline path)
- `fleet_status(format="json")` — already had `version` field, unchanged

**Usage:** Call any tool with `format="json"` and `server_version` appears at the top of the response.


kumaakh:
Merged into integration/schema-oauth

PR #155, opened 2026-04-18
fix(install): rename installer artifact to avoid collision with installed binary

kumaakh:
## Problem

When a user downloaded `apra-fleet.exe` (or `apra-fleet` on Unix) and ran `apra-fleet.exe install`, the running-process check detected the installer itself in the process list and falsely reported the server was already running — forcing users to add `--force` even when nothing was actually running.

### 1. Rename release artifacts (`package-sea.mjs`, `ci.yml`, docs)

All release binaries are now named `apra-fleet-installer-{platform}-{arch}[.exe]`:

| Before | After |
|--------|-------|
| `apra-fleet-win-x64.exe` | `apra-fleet-installer-win-x64.exe` |
| `apra-fleet-linux-x64` | `apra-fleet-installer-linux-x64` |
| `apra-fleet-darwin-arm64` | `apra-fleet-installer-darwin-arm64` |

The downloaded installer name never collides with the installed `~/.apra-fleet/bin/apra-fleet[.exe]` binary.

### 2. Harden `isApraFleetRunning()` on Windows (`src/cli/install.ts`)

Switches `tasklist` to CSV output format and excludes the current process PID, so self-updates (where the installed `apra-fleet.exe` itself runs `install`) work correctly without `--force`.

## CI / signing

- Matrix `binary` names updated → `package-sea.mjs` produces the matching filename
- `sign-windows` download/upload artifact names updated — file lands at `to-sign/apra-fleet-installer-win-x64.exe`, signed in-place, re-uploaded under the same name ✓
- Release `pattern: apra-fleet-*` still matches `apra-fleet-installer-*` ✓
- `llms-full.txt` intentionally not changed here — CI regenerates it from `README.md` on this PR

## Test plan
- [ ] Fresh install: download `apra-fleet-installer-win-x64.exe`, run `.\apra-fleet-installer.exe install` — should succeed without `--force`
- [ ] Self-update: run `apra-fleet install` from installed binary while server is stopped — should succeed
- [ ] Blocked correctly: run `apra-fleet install` while server is actually running — should still show the error

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,11 +109,11 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            binary: apra-fleet-linux-x64
+            binary: apra-fleet-installer-linux-x64
           - os: macos-latest
-            binary: apra-fleet-darwin-arm64
+            binary: apra-fleet-installer-darwin-arm64
           - os: windows-latest
-            binary: apra-fleet-win-x64.exe
+            binary: apra-fleet-installer-win-x64.exe
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout repository
@@ -191,7 +191,7 @@ jobs:
       - name: Download Windows binary
         uses: actions/download-artifact@v4
         with:
-          name: apra-fleet-win-x64.exe
+          name: apra-fleet-installer-win-x64.exe
           path: to-sign
 
       - name: Azure Login (OIDC)
@@ -216,8 +216,8 @@ jobs:
       - name: Upload signed binary
         uses: actions/upload-artifact@v4
         with:
-          name: apra-fleet-win-x64.exe
-          path: to-sign/apra-fleet-win-x64.exe
+          name: apra-fleet-installer-win-x64.exe
+          path: to-sign/apra-fleet-installer-win-x64.exe
           retention-days: 30
           overwrite: true
 
@@ -311,9 +311,9 @@ jobs:
         with:
           files: |
             ${{ env.TARBALL_NAME }}
-            release-binaries/apra-fleet-linux-x64
-            release-binaries/apra-fleet-darwin-arm64
-            release-binaries/apra-fleet-win-x64.exe
+            release-binaries/apra-fleet-installer-linux-x64
+            release-binaries/apra-fleet-installer-darwin-arm64
+            release-binaries/apra-fleet-installer-win-x64.exe
           generate_release_notes: true
 
       - name: Bump version.json via PR
--- a/README.md
+++ b/README.md
@@ -62,17 +62,17 @@ Copy-paste the one-liner for your platform:
 
 **macOS (Apple Silicon)**
 ```bash
-curl -fsSL https://github.com/Apra-Labs/apra-fleet/releases/latest/download/apra-fleet-darwin-arm64 -o apra-fleet && chmod +x apra-fleet && ./apra-fleet install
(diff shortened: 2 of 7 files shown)


kumaakh (commented):
## Review: fix(install): rename installer artifact to avoid collision

**Overall**: Good fix for a real UX problem. The artifact rename and CI plumbing look correct. However, there are issues that need addressing before merge.

---

### 🔴 Tests not updated for new CSV-based `isApraFleetRunning` (blocking)

The Windows path in `isApraFleetRunning()` now uses `/FO CSV` output and regex parsing, but the existing tests in `tests/install-force.test.ts:223-232` still mock the **old plain-text** format:

```ts
// Line 225 — still mocking old format:
vi.mocked(execSync).mockReturnValue('apra-fleet.exe  1234 Console' as any);
```

This causes **3 test failures** (`npm test -- --run tests/install-force.test.ts`). The mocks need to return CSV-formatted output, e.g.:

```ts
vi.mocked(execSync).mockReturnValue('"apra-fleet.exe","5678","Console","1","14,000 K"' as any);
```

You should also add a test for the PID-exclusion logic — mock the current `process.pid` and verify that a tasklist line containing that PID returns `false` while a different PID returns `true`.

---

### 🟡 Incorrect comment about `pgrep` (minor)

```ts
// -x = exact name match; pgrep excludes the calling process by default (NOTE 2)
```

`pgrep` does **not** exclude the calling process by default. It only excludes itself (the `pgrep` process). The reason the installer doesn't self-match on Unix is that it's now named `apra-fleet-installer-*`, so `-x apra-fleet` won't match it. The comment should reflect that:

```ts
// -x = exact name match; the installer binary is named apra-fleet-installer-*, so it won't match
```

Note: in the **self-update** scenario (running `~/.apra-fleet/bin/apra-fleet install`), the process IS named `apra-fleet` and pgrep WILL match it — same class of bug you fixed for Windows. Consider applying the same PID-exclusion approach on Unix too, or document it as a known limitation.

---

### ✅ What looks good

- Artifact rename in `package-sea.mjs`, `ci.yml`, and release step is consistent
- `sign-windows` job correctly references the new artifact names
- `release-binaries/apra-fleet-*` glob still matches `apra-fleet-installer-*` ✓
- README and deploy.md docs updated consistently
- Windows `isApraFleetRunning` logic (CSV parsing + PID exclusion) is the right approach

kumaakh (commented):
## Re-review: All prior issues addressed ✓

Both issues from my previous review have been resolved in the latest push (`69aaeb1`):

### 1. Tests updated for CSV-based `isApraFleetRunning` ✓ (was blocking)

- `mockServerRunning()` and the Windows `--force` test now return proper CSV format (`"apra-fleet.exe","5678","Console","1","14,000 K"`)
- **Two new PID-exclusion tests added** — one for Linux (`pgrep` returns only current PID → `false`) and one for Windows (`tasklist` CSV contains only current PID → `false`). Good coverage.
- All 16 tests in `install-force.test.ts` pass. Full suite: 750/750 pass.

### 2. Comment corrected + Unix PID-exclusion added ✓ (was minor)

- Comment now correctly states: `installer is apra-fleet-installer-* so won't match` — accurate.
- **Bonus**: Unix `isApraFleetRunning()` now also applies PID-exclusion via `pgrep` output parsing, handling the self-update scenario (`~/.apra-fleet/bin/apra-fleet install`) I flagged. Same approach as the Windows fix — clean and consistent.

### 3. `llms-full.txt` updated ✓

The regenerated `llms-full.txt` is included and matches the README changes.

LGTM — ready to merge.

Issue #193, opened 2026-04-27
enhancement: per-instance data directory isolation for multi-fleet on same device


kumaakh:
## Problem

Multiple fleet instances running on the same device all share the same data directory (`~/.apra-fleet/data/`). This causes:

- **Race conditions** on `registry.json`, `statusline.txt`, `known_hosts`, `salt` — two instances writing simultaneously
- **Registry cross-contamination** — instance A sees instance B's members and vice versa
- **Statusline collisions** — last writer wins, statusline reflects whichever instance wrote most recently

Common scenario: a developer running separate fleet instances for different projects (e.g. one for `apra-fleet` development, one for `odm` work) on the same machine.

## Current State

`APRA_FLEET_DATA_DIR` env var already exists and is fully respected:

- `src/paths.ts`: `export const FLEET_DIR = process.env.APRA_FLEET_DATA_DIR ?? path.join(os.homedir(), '.apra-fleet', 'data')`
- `scripts/fleet-statusline.sh`: `STATUSLINE_FILE="${APRA_FLEET_DATA_DIR:-$HOME/.apra-fleet/data}/statusline.txt"`
- Tests already use it: `process.env.APRA_FLEET_DATA_DIR = path.join(os.tmpdir(), 'apra-fleet-test-data')`

The mechanism is complete — it just isn't surfaced as a first-class configuration option at install/registration time.

### 1. `apra-fleet install --data-dir <path>` flag

Allow specifying a custom data directory at install time. The installer writes `APRA_FLEET_DATA_DIR=<path>` into the MCP server's `env` config so it is passed automatically on every server start.

Example:
```bash
apra-fleet install --data-dir ~/.apra-fleet/projects/odm
apra-fleet install --data-dir ~/.apra-fleet/projects/apra-fleet
```

Each installation gets its own isolated registry, statusline, known_hosts, and salt.

### 2. MCP registration env field

For Claude Code, the `env` field in the MCP config carries the variable through:

```json
{
  "mcpServers": {
    "apra-fleet-odm": {
      "command": "apra-fleet",
      "env": { "APRA_FLEET_DATA_DIR": "/Users/akhil/.apra-fleet/projects/odm" }
    }
  }
}
```

This allows multiple named fleet MCP registrations on the same machine, each with isolated data.

### 3. Named instances (optional, builds on above)

[...]

PR #231, opened 2026-05-03
feat(install): per-instance data dir isolation via --data-dir / --instance (#193)

yashrajsapra:
## Summary

- Adds `--data-dir` and `--instance` flags to `install` so multiple Claude Code instances running on the same machine use isolated config/log directories instead of fighting over a shared one
- Shell injection in `--data-dir` is eliminated by switching to `execFileSync` argv array (security fix surfaced in review)
- Multi-instance usage guide added to docs

## Changes

- `feat(install)`: per-instance data dir isolation via `--data-dir` / `--instance` + workspace CLI
- `docs`: multi-instance usage guide
- `fix(install)`: eliminate shell injection in `--data-dir` via `execFileSync` argv array
- `chore`: cleanup fleet control files (PLAN.md, progress.json, feedback.md, CLAUDE.md, AGENTS.md)

## Test plan

- [ ] `node dist/index.js install --data-dir /tmp/inst-a` and `--data-dir /tmp/inst-b` produce fully isolated runtime dirs
- [ ] `node dist/index.js install --instance myproject` resolves to a deterministic subdirectory
- [ ] Passing a path with spaces / special chars in `--data-dir` does not cause shell injection
- [ ] Existing installs without `--data-dir` continue to use the default data directory unchanged
- [ ] Multi-instance usage guide renders correctly in docs


src/cli/install.ts
```diff
@@ -541,19 +592,42 @@ ${killHint}
 
   if (llm === 'claude') {
     try {
-      run('claude mcp remove apra-fleet --scope user', { stdio: 'ignore' });
+      run(`claude mcp remove ${serverName} --scope user`, { stdio: 'ignore' });
```
kumaakh, line 595: **MEDIUM (non-blocking):** `mcp remove` still goes through `run()`, which calls `execSync` with the command as a shell string. `serverName` is validated upstream so this is safe in practice, but it is inconsistent with the `execFileSync` argv-array fix applied everywhere else in this PR.

Consider splitting `run()` into a `runFile(bin, args[])` variant, or inlining this call as:
```ts
execFileSync('claude', ['mcp', 'remove', serverName, '--scope', 'user'], { stdio: 'ignore' });
```
This keeps the security posture consistent across the whole install path.

src/cli/workspace.ts
```diff
@@ -0,0 +1,281 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import os from 'node:os';
+
+const home = os.homedir();
+const APRA_BASE = path.join(home, '.apra-fleet');
```
kumaakh, line 6: **LOW (non-blocking):** `APRA_BASE` and `DEFAULT_DATA_DIR` are re-derived here from scratch, but `paths.ts` already exports `APRA_BASE` and `FLEET_DIR` (which is `APRA_FLEET_DATA_DIR ?? .apra-fleet/data`). Importing from `paths.ts` would eliminate the duplication and ensure both files always agree on the canonical base path — especially important if the default ever changes.

src/cli/install.ts
```diff
@@ -598,7 +672,7 @@ ${killHint}
   }
 
   // Finalize permissions
-  mergePermissions(paths);
+  mergePermissions(paths, serverName);
 
   // Write install-config.json
   const installConfig = { llm, skill: skillMode };
```
kumaakh, line 678: **BLOCKING** — `installConfig` omits `dataDir` and `instanceName`. When `apra-fleet update` reads this file to replay the install, it has no record of the custom data dir and will reinstall for the default workspace only, silently breaking any non-default workspace install. See review body for the full issue and suggested fix.

kumaakh (commented):
**BLOCKING — `apra-fleet update` silently loses custom workspace on replay**

Two issues in this section mean that `apra-fleet update` will always reinstall to the *default* workspace, silently discarding any `--data-dir` / `--instance` the user originally passed.

**Issue 1 (line 678):** `installConfig` only saves `{ llm, skill }`. `dataDir` and `instanceName` are not persisted.

**Issue 2 (line 679):** `configDir` is hardcoded to `FLEET_BASE/data` (i.e. `~/.apra-fleet/data/`). When a custom `--data-dir` is used, the config is written to the *wrong* directory — the default workspace's data folder instead of the custom one.

**Issue 3 (not in this diff — `src/cli/update.ts` line 91):** Even if `installConfig` were fixed to save `dataDir`, the update replay only constructs `['install', '--llm', config.llm, '--skill', config.skill]` — `--data-dir` and `--instance` are never read back or passed.

**Fix needed in this PR:**
```typescript
// line 678 — include dataDir / instanceName
const installConfig: Record<string, string> = { llm, skill: skillMode };
if (dataDir) installConfig.dataDir = dataDir;
if (instanceName) installConfig.instance = instanceName;

// line 679 — write config to the active data dir, not hardcoded default
const configDir = dataDir ?? path.join(FLEET_BASE, 'data');
```
And `update.ts` must be updated to read `config.dataDir` / `config.instance` and append `--data-dir` / `--instance` to the replay args.

kumaakh:
**Design suggestion for the `install-config.json` fix**

Rather than a per-workspace file, `install-config.json` should be a **singleton accumulator** at `~/.apra-fleet/install-config.json` (outside any data dir) that tracks every install invocation:

```json
{
  "installs": [
    { "llm": "claude", "skill": "all" },
    { "llm": "claude", "skill": "fleet", "instance": "work" },
    { "llm": "gemini", "skill": "fleet", "dataDir": "/custom/path" }
  ]
}
```

Each entry maps 1:1 to a CLI invocation. `apra-fleet update` reads this one file and replays every entry in order — no need to discover workspaces via `workspaces.json` first.

**Upsert rule:** key on `{ instance, dataDir }` (both optional/undefined for the default workspace). On re-install, replace the matching entry rather than append — so running `apra-fleet install --instance work --llm gemini` a second time updates that entry in place.

**Changes needed:**
- `src/cli/install.ts` — after MCP registration, upsert into `~/.apra-fleet/install-config.json` (not `${dataDir}/install-config.json`)
- `src/cli/update.ts` — read `config.installs[]`, replay each entry with `['install', '--llm', e.llm, '--skill', e.skill, ...(e.instance ? ['--instance', e.instance] : []), ...(e.dataDir ? ['--data-dir', e.dataDir] : [])]`
- Tests: add a case that installs twice with different `--instance` values and verifies `update` would replay both

PR #274, opened 2026-05-20
Consume blindfold as a dependency; remove in-tree security code

mraduldubey:
## Summary

- Extracts the credential-security system (auth-socket, credential-store, crypto, shell-escape, OOB collection, token resolver) into the standalone [blindfold](https://github.com/Apra-Labs/blindfold) package and consumes it as a dependency. Fleet no longer maintains the security layer in-tree.
- Moves the egress-confirmation subcommand from `apra-fleet secret --confirm` to `apra-fleet auth --confirm` (matching what blindfold's OOB launcher spawns). The old path is removed entirely.
- Hardens test isolation: `vitest.config.ts` + `tests/setup.ts` now make it impossible for `npm test` to write to the real `~/.apra-fleet/data` registry (root-caused a load-order bug that had been silently polluting the live registry during test runs).
- Consumes **`@apralabs/blindfold@^0.1.1`** from the npm registry. (0.1.x carries the v0.0.2 hardening below - `resolve_secure` removed, vault-only MCP surface - plus the gnome-terminal `--wait` OOB credential-entry fix.)

## What changed under the hood

- `package.json` pulls blindfold from the npm registry via an alias: `"blindfold": "npm:@apralabs/blindfold@^0.1.1"`. The alias keeps the dependency key `blindfold`, so every `from 'blindfold'` import is unchanged. (No git submodule - the earlier `file:./blindfold` submodule approach was replaced once blindfold was published to npm.)
- New `src/services/blindfold-init.ts` calls `initBlindfold({ dataDir: FLEET_DIR, productName: 'apra-fleet', pipeName: 'apra-fleet-auth' })` at every entrypoint after `--version`/`--help` short-circuit. This preserves existing users' on-disk state (`~/.apra-fleet/data/credentials.json`, the auth socket path, and the Windows named pipe name) - no migration required.
- All ~26 fleet source files + ~11 test files now import security primitives via `from 'blindfold'`. Fleet-local re-implementations of `resolveSecureTokens` / `redactOutput` / `resolveSecureField` / `SECURE_TOKEN_RE` are removed in favor of blindfold's canonical exports.
- 9 src files and 7 test files deleted (covered by blindfold/tests/).
- No build-time scaffolding: blindfold installs prebuilt from the registry (`node_modules/blindfold` -> `@apralabs/blindfold`). The submodule-building `postinstall`, the `blindfold-ci.yml` workflow, and the `submodules: recursive` checkout steps (ci.yml x4, fleet-e2e.yml) are all removed.

## Blindfold security hardening (v0.0.2)

The `resolve_secure` MCP tool that v0.0.1 shipped is a fundamental hole in blindfold's threat model: it took a string containing `{{secure.NAME}}` tokens and returned the same string with plaintext credentials substituted in - directly to the LLM caller. Even with redaction markers in the response, the plaintext bytes were already in the LLM's context window, transcript, and any upstream logging.

v0.0.2 removes that tool. The blindfold MCP wire surface is now exactly four tools, all vault-management only:

- `credential_store_set` - OOB collection, returns a handle
- `credential_store_list` - metadata only
- `credential_store_update` - metadata only
- `credential_store_delete` - returns status

**Library exports are unchanged** (`resolveSecureTokens`, `resolveSecureField`, `redactOutput`, `containsSecureTokens`, `SECURE_TOKEN_RE`, `SEC_HANDLE_RE`). apra-fleet uses those server-side inside `execute_command` to substitute tokens just before shell exec and redact plaintext from output before returning to the LLM. Plaintext therefore continues to never cross the LLM boundary in apra-fleet workflows.

Implication for standalone blindfold users: blindfold-without-a-host now stores credentials but cannot resolve them in agentic workflows. That is the intended design - vault-only. Hosts (apra-fleet, or anyone consuming the library) own the execution path and the secret-aware redaction. Documented in blindfold's `README.md` under "Standalone vs host-integrated usage".

## Compatibility

- MCP tool surface (`credential_store_*`, `execute_command`, etc.): unchanged names, schemas, and response shapes.
- CLI surface: `secret --set/--list/--update/--delete` unchanged. Only `secret --confirm` moved to `auth --confirm`.
- Existing users' credentials, sockets, and pipes: preserved bit-for-bit through `initBlindfold` configuration.
- blindfold's removed `resolve_secure` MCP tool: not consumed by apra-fleet (it always used the library function), so zero behavioral impact on apra-fleet.

## Test plan

- [x] `npm run build` (tsc) - clean
- [x] `npm test` - 1169 passing, 3 failing (all pre-existing baseline: 1 `platform.test.ts` login-shell env probe, 2 `time-utils.test.ts` IST timezone). Independently re-verified by reviewer in a separate clone.
- [x] `npm run smoke` - 13/13 pass with isolated data dir
- [x] `npm run build:binary` - SEA produces a binary that boots `--version` in under 400 ms across linux/macos/windows matrix
- [x] CLI round-trip in an isolated `APRA_FLEET_DATA_DIR` - `secret --set` / `--list` / `--update` / `--delete` all green
- [x] Test-isolation hardening verified empirically - snapshot-diff of `~/.apra-fleet/data/registry.json` across multiple `npm test` runs returns 0 lines
- [x] Clean install verified - `rm -rf node_modules && npm install` resolves `blindfold@npm:@apralabs/blindfold@0.1.1` from the registry; `npm run build` clean; suite green (1197 passing; same 3 pre-existing env failures - 1 login-shell probe, 2 IST timezone)
- [x] `apra-fleet auth --confirm <bad name>` rejects with exit 1
- [x] blindfold MCP wire surface inspected via live `tools/list` JSON-RPC probe - exactly 4 vault tools, no `resolve_secure`
- [x] blindfold's own test suite: 139 passing, 0 failing
- [x] ASCII-only sprint-wide, no AI / Claude / Anthropic attribution traces

## Direct-use testing (NOT done in this PR)

These were deliberately skipped to keep the PR isolated from the reviewer's live MCP setup. Recommended before merge:

1. Point a separate Claude Code session at `node /path/to/dist/index.js` with a custom `APRA_FLEET_DATA_DIR`; drive `register_member` + `credential_store_set` + `execute_command echo {{secure.FOO}}` (verify `[REDACTED:FOO]` in output).
2. Set a credential with `network_policy=confirm`, run a curl-style `execute_command`, verify an OOB terminal pops running `apra-fleet auth --confirm`.
3. `provision_vcs_auth` for a test repo to exercise the `collectOobApiKey` path.

## npm migration (done)

blindfold is now published as `@apralabs/blindfold` on npm. This PR consumes
`@apralabs/blindfold@^0.1.1` via the npm alias and drops the git submodule entirely:
`.gitmodules`, the submodule gitlink, the `postinstall` builder, `blindfold-ci.yml`, and the
`submodules: recursive` CI checkout steps are all removed. The lockfile pins the exact
resolved tarball from `registry.npmjs.org`.


kumaakh:
@mraduldubey pls rebase and resolve these conflicts so that we can close this PR... thanks

PR #278, opened 2026-05-28
ci(llms-full): regenerate on push:main via apra-fleet-git app

kumaakh:
## Summary

Moves the llms-full.txt auto-regen from the PR path to a post-merge push:main path. Authenticates via the apra-fleet-git GitHub App which is configured as an Integration bypass actor in the new main-protection ruleset, so the regen commit can push directly to main without needing required status checks.

## Why

The previous implementation auto-committed regenerated llms-full.txt back to PR branches as github-actions[bot]. That:
- Advanced the PR head past the original commit
- The new commit only touched llms-full.txt (in paths-ignore) so CI did not re-run
- Required status checks never registered on the PR head, blocking merge

Result: every PR with a non-trivial change ended up requiring an empty-commit workaround or admin override to merge. Confirmed during the recent feat/agy-support merge.

## What changed

- update-llms-full job now triggers only on push:main (not pull_request)
- Authenticates with the apra-fleet-git GitHub App (id 3001109) via actions/create-github-app-token
- Pushes the regenerated file directly to main; bypass is granted by the new ruleset

## One-time setup (needed before this PR is useful after merge)

1. Repo variable: APRA_FLEET_GIT_APP_ID = 3001109
2. Repo secret: APRA_FLEET_GIT_APP_PRIVATE_KEY = PEM private key downloaded from https://github.com/organizations/Apra-Labs/settings/apps/apra-fleet-git

The first push:main after merging this PR will exercise the new path. If the secret/var are missing the workflow fails visibly at the Mint app token step.

## Test plan
- [x] CI build-and-test passes on this branch (will be a required check before merge)
- [ ] After merge: confirm update-llms-full job runs and either no-ops (if llms-full.txt already current) or commits + pushes successfully

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -229,17 +229,28 @@ jobs:
           overwrite: true
 
   update-llms-full:
+    # Runs only on push to main (i.e. after a PR has merged). Bypasses the
+    # main-protection ruleset via the apra-fleet-git GitHub App, which is
+    # configured as a bypass actor. No-op when llms-full.txt is already current.
+    # Requires: secret APRA_FLEET_GIT_APP_PRIVATE_KEY (PEM) and variable
+    # APRA_FLEET_GIT_APP_ID.
     needs: build-and-test
-    if: github.event_name == 'pull_request'
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


kumaakh:
Superseded by the pre-commit-hook + CI-check approach. The bypass route added too much attack surface for the actual requirement.

PR #319, opened 2026-07-05
feat(watch): live member activity viewer

### LINKED ISSUE #318: feat: live member activity viewer (apra-fleet watch)
## Problem

A dispatched member is a black box. Today the only ways to see what a member is
doing are:

- `execute_prompt` -- returns a single final blob, nothing while it runs
- `execute_command` + `monitor_task` -- pollable but shell-only and opt-in

Neither lets a human simply *watch* members work, and neither lets you watch
several in parallel -- which is the whole point of a fleet. There is no
`docker compose logs -f` for the fleet.

## Proposal

Add `apra-fleet watch`: a standalone CLI subcommand that streams live member
activity to the terminal, multiplexed and grouped, with no MCP connection
required.

Merge two sources per member:

| Source | Captures | Members |
|--------|----------|---------|
| Fleet activity log (`FLEET_DIR/logs/fleet-<pid>.log`) | every dispatch: shell commands + output, prompt invocations, file transfers | all (local + remote) |
| Provider transcript (`~/.claude/projects/*.jsonl`) | LLM reasoning, edits, tool output | local only |

The fleet log is the universal spine (works for remote members with no ssh2
round-trip, since the server logs at dispatch time); the transcript is
local-only enrichment layered in for local Claude members.

## Scope

Three zoom levels, no "track" jargon:
- project = git origin (worktrees/clones group together)
- feature = git branch
- member = name

Command surface:
- `watch` -- follow members, scope inferred from cwd
- `watch <name>...` -- follow specific members
- `watch --project <dir>` / `--feature <name>` / `--branch <ref>` -- scoped follow
- `watch --list` -- overview and exit
- `watch --tail <n>` -- backfill last n events
- `watch --verbose|-v` -- edit diffs, file contents, command output, thinking

Marker-forward, colorized rendering (`>` read-only/dispatch, `*` edits,
`$` shell, dim lifecycle). Follow model streams present + future activity with
no activity gate (idle members stream the moment they produce output).

## Out of scope (deferred)

- Gemini / other-provider transcript parsing (fleet-log activity still works;
  only the rich LLM detail is Claude-only)
- Push-based SSE/WebSocket sink and a multi-pane web/TUI
- PM `status.md` feature-name labeling (branch names shown in the interim)

vbv-nyk:
Closes #318

## What

Adds `apra-fleet watch` -- a `docker compose logs -f` for the fleet. Streams
live member activity to the terminal, multiplexed and grouped, no MCP
connection required.

## How it works

Merges two sources per member:

| Source | Captures | Members |
|--------|----------|---------|
| Fleet activity log (`FLEET_DIR/logs/fleet-<pid>.log`) | every dispatch: shell commands + output, prompt invocations, file transfers | all (local + remote) |
| Provider transcript (`~/.claude/projects/*.jsonl`) | LLM reasoning, edits, tool output | local only |

The fleet log is the universal spine -- the server records every dispatch
locally at dispatch time, so remote members are observable with no ssh2
round-trip. The transcript is local-only enrichment layered in for local Claude
members. Scope resolves to a member set via git (project = origin, feature =
branch), then log lines are gated by member `mid`/`mem` attribution.

## Command surface

    apra-fleet watch                     Follow members (scope inferred from cwd)
    apra-fleet watch <name>...           Follow specific members by name
    apra-fleet watch --project <dir>     Follow members on the repo at <dir>
    apra-fleet watch --feature <name>    Follow members on one feature (branch)
    apra-fleet watch --branch <ref>      Follow members on an exact branch
    apra-fleet watch --list              Print the overview and exit
    apra-fleet watch --tail <n>          Backfill the last n events per member
    apra-fleet watch --verbose | -v      Edit diffs, contents, output, thinking

## Changes

- `src/cli/watch.ts` -- the subcommand: scope resolution, overview, 700ms poll
  loop tailing both sources by byte-offset, colorized marker-forward rendering,
  fleet-log rollover handling (strictly-newer mtime, jump-to-EOF)
- `src/services/watch/fleet-log.ts` -- newest-log resolution, per-line JSONL
  parse + member attribution, noise-tag filtering, recent-activity scan
- `src/services/watch/project-resolver.ts` -- git enrichment, origin
  normalization, worktree-suffix stripping, project/feature grouping
- `src/services/watch/transcript-formatter.ts` -- Claude transcript -> display
  lines (prose, tool headers, verbose diffs/content/output/thinking)
- `src/tools/execute-command.ts` -- log command output under a `command_output`
  tag (capped) so short-command stdout surfaces in `watch`
- `src/tools/execute-prompt.ts` + `src/utils/log-helpers.ts` -- `singleLineForLog`
  drops the 80-char truncation so full commands/prompts appear in the stream
- `src/index.ts` -- wire up the `watch` subcommand
- `docs/features/watch.md` -- feature documentation

## Testing

- Watch unit tests all pass (37 tests): `tests/watch-fleet-log.test.ts`,
  `tests/watch-project-resolver.test.ts`, `tests/watch-transcript-formatter.test.ts`
  (parsing, grouping, worktree-stripping, noise filtering, formatting)
- Note: `tests/time-utils.test.ts` shows 2 failures in a non-UTC local
  timezone; they are pre-existing, unrelated to this change (this branch does
  not touch `time-utils`), and pass under `TZ=UTC`.

## Deferred

- Gemini/other-provider transcript parsing (fleet-log activity still works)
- Push-based SSE/WebSocket sink and a multi-pane web/TUI
- PM `status.md` feature-name labeling (branch names shown in the interim)


kumaakh:
## Feedback: log verbosity vs. persisted sensitive data

Nice feature, and the fleet-log architecture (universal spine + local transcript enrichment) is a clean way to make remote members observable without ssh2. Ran the full suite locally (1630/1630 passing) and it's functionally solid.

One nuance worth discussing before merge: `singleLineForLog` (replacing `truncateForLog`'s 80-char cap) means `execute_command`/`execute_prompt` now write the **full, uncapped** command/prompt text into the fleet log, instead of an 80-char snippet. `maskSecrets` still runs first, so known credential patterns (`{{secure.NAME}}`, `sec://...`) stay redacted -- but anything that doesn't match those patterns (inline API keys, internal URLs, customer data pasted into a prompt, etc.) now persists to disk in full rather than as a truncated snippet. It also removes the implicit cap on fleet-log growth per dispatch (command_output already has a cap via `capForLog`; the command/prompt line itself doesn't).

A middle-ground alternative: keep the primary `LogScope` line truncated at 80 chars (unchanged behavior for anything else parsing that line), and emit the full command/prompt on a **separate** log tag (e.g. `command_full` / `prompt_full`) -- same pattern already used for `command_output` today. `watch` would read that tag for full-text rendering, while the original summary line stays short. I'd still cap that new tag generously (a few thousand chars, mirroring `capForLog`) rather than leaving it fully unbounded.

Curious what you think -- is the full-text-in-primary-line tradeoff intentional/acceptable for this log's retention model, or would the separate-tag approach be worth doing before merge?

vbv-nyk-apra:
Pushed 3 follow-up commits extending the viewer to remote members and rebalancing the default view.

### 1. Stream remote member transcripts over SSH (\`feat\`)
Previously \`watch\` could only show a remote member's shell dispatches (from the fleet log), never its LLM session detail -- that transcript lives on the member's own disk. Now, for remote Claude members, it is tailed over a **dedicated, long-lived \`tail -F\` SSH channel** (push, not poll):
- \`ssh.ts\`: new \`execStream()\` -- a non-pooled streaming exec for long-lived commands, so the request/response pool's idle timer never tears the tail down.
- \`watch.ts\`: \`ensureRemoteTail()\` opens the tail on the newest session \`.jsonl\`. First attach primes to EOF (\`-n0\`, no history dump); a session rotation reopens from the top (\`-n +1\`). A light periodic \`ls\` only reopens a dead channel or follows a rotation. Channels close cleanly on Ctrl-C; failures fail soft and retry.
- Local and remote transcripts feed the **same** \`formatTranscriptLine\` renderer.

### 2. Show tool output/diffs by default, reserve \`-v\` for thinking (\`refactor\`)
The verbosity split was backwards -- tool results (the actual logs) were \`-v\`-only. Now command/tool output, edit diffs, \`Write\` contents, and multi-line command bodies render **by default**; \`Edit\`/\`Write\` headers carry a size summary (\`Edit foo.ts (+3 -1)\`). **Thinking is the only content still gated behind \`-v\`.**

### 3. Fix Claude session-dir path encoding (\`fix\`)
The resolver used \`[\/\\:] -> -\`, but Claude's real rule is \`[^a-zA-Z0-9] -> -\`. A path like \`/home/ecs_user/...\` resolved to \`-home-ecs_user-...\` instead of \`-home-ecs-user-...\`, so \`watch\` (and the stall detector) silently found no transcript for any path containing \`_\`/\`.\`/spaces. Factored into a shared \`encodeClaudeProjectDir()\`.

**Verification:** built clean; full suite passes (excluding a pre-existing timezone test in \`time-utils.test.ts\`, untouched here). Confirmed live against a remote member -- reasoning, tool calls, and command output stream in real time. \`docs/features/watch.md\` updated to match.

kumaakh:
@vbv-nyk Re-reviewed after the 3 new commits (`00802fa`, `0e4e648`, `4e54a1d`) -- now 16 files / +1587. Ran a full build + test pass again: clean build, **1633 passed / 14 skipped** (up from 1630, consistent with the new coverage added).

### `4e54a1d` (session-dir path-encoding fix)
Correct and well-scoped -- `encodeClaudeProjectDir` now replaces *every* non-alphanumeric char with `-`, matching Claude Code's actual on-disk directory-naming convention (previously only `[\/\:]` were handled, missing `_`/`.`). Good regression test added in `tests/log-path-resolver.test.ts` for the underscore case.

### `0e4e648` (SSH remote transcript streaming) -- the highest-risk addition
- **Injection safety looks solid.** The remote directory component comes from `encodeClaudeProjectDir`, whose output charset is restricted to `[a-zA-Z0-9-]` by construction, so it's safe to interpolate unquoted. The untrusted part -- the remote filename returned by `ls -t` -- is correctly passed through a `shq()` helper before landing in the `tail` command, using the same escaping approach as `escapeShellArg` in `src/utils/shell-escape.ts`.
- **Credentials:** `execStream` reuses the existing `getSSHConfig(agent)` resolution path (same key/password + known-hosts handling as pooled `execCommand`) -- nothing new embedded in constructed commands or logged.
- **Two non-blocking follow-ups:**
  1. `shq()` in `watch.ts` duplicates `escapeShellArg` instead of importing the shared one from `shell-escape.ts` -- functionally fine, but worth consolidating so there's one escaping implementation to audit.
  2. No unit tests directly exercise `execStream`/`ensureRemoteTail` with an adversarial filename (e.g. containing `; rm -rf` or a stray quote). Given this is new remote-access surface, a test asserting `shq()` is actually applied would be good insurance even though the current escaping reads correct.

### `00802fa` (verbosity refactor)
Docs (`docs/features/watch.md`) and `printUsage()` help text were updated consistently with the new `-v` semantics ("thinking is the only content reserved for -v"), and `tests/watch-transcript-formatter.test.ts` was rewritten to match. No stale doc/test drift found.

---

### Highlighting my earlier comment -- still open, not addressed by these 3 commits
The [log-retention question from my previous review](#issuecomment-4888855462) is unrelated to this batch of commits and still stands: `singleLineForLog` (replacing the old `truncateForLog(text, 80)` cap) means `execute_command`/`execute_prompt` now persist the **full, uncapped** (masked-secrets-only) command/prompt text to the fleet log instead of an 80-char snippet. Proposed middle ground was: keep the primary log line truncated as before, and add the full text under a separate capped tag (e.g. `command_full`/`prompt_full`, mirroring how `command_output` already works) for `watch` to render.

@vbv-nyk would appreciate your take on that tradeoff before merge -- otherwise this PR looks ready to me (build clean, 1633/1633 non-skipped tests passing, no injection issues found in the new SSH path).

kumaakh (commented):
Inline follow-up to my two prior review comments -- pinning each point to its exact location for easier tracking. @vbv-nyk

src/cli/watch.ts
```diff
@@ -0,0 +1,530 @@
...
+  for (const line of lines) {
+    for (const ev of formatTranscriptLine(f.provider, line, verbose)) collected.push(ev);
+  }
+  const toPrint = tailN > 0 && collected.length > tailN ? collected.slice(-tailN) : collected;
+  for (const ev of toPrint) emit(f, ev, single);
+}
+
+/** Single-quote a string for safe interpolation into a bash command. */
+function shq(s: string): string {
+  return `'${s.replace(/'/g, `'\\''`)}'`;
```
kumaakh: `shq()` duplicates the single-quote escaping algorithm already implemented as `escapeShellArg` in `src/utils/shell-escape.ts:11`. Not a vulnerability -- the logic is equivalent -- but worth importing the shared helper instead of a second copy, so there's one escaping implementation to audit/maintain.

src/cli/watch.ts
```diff
@@ -0,0 +1,530 @@
...
+ *   - first open primes to EOF (`-n0`) so no history is dumped;
+ *   - a rotation opens the new session from its top (`-n +1`) so nothing is missed.
+ * Fails soft: connect/exec errors leave rtStream null and the next check retries.
+ */
+async function ensureRemoteTail(f: Follower, single: boolean, verbose: boolean): Promise<void> {
+  if (!f.rtEnc || f.rtBusy) return;
+  f.rtBusy = true;
+  try {
+    const dir = `"$HOME/.claude/projects/${f.rtEnc}"`;
+    const res = await execCommand(f.agent, `ls -t ${dir}/*.jsonl 2>/dev/null | head -1`, 8000);
```
kumaakh, line 427: `newest` (the remote filename from `ls -t`) is untrusted input and is correctly passed through `shq()` before being interpolated into the `tail` command below (line 431). Confirmed this is injection-safe. One gap: no unit test exercises `ensureRemoteTail`/`execStream` with an adversarial filename (e.g. containing `; rm -rf` or a stray `'`) to lock in that `shq()` is actually applied here. Given this is new remote-access surface, a regression test here would be good insurance.

src/tools/execute-command.ts
```diff
@@ -122,7 +132,7 @@ export async function executeCommand(input: ExecuteCommandInput, extra?: any): P
   }
 
   const strategy = getStrategy(agent);
-    const scope = new LogScope('execute_command', `${truncateForLog(maskSecrets(input.command))}`, agent);
+    const scope = new LogScope('execute_command', `${singleLineForLog(maskSecrets(input.command))}`, agent);
```
kumaakh: This is the log-retention point from my earlier PR comment, pinned to its exact line: swapping `truncateForLog(.., 80)` for `singleLineForLog` means the **full** (masked-secrets-only) command text is now persisted to the fleet log, not an 80-char snippet. `maskSecrets` only catches known credential patterns (`{{secure.NAME}}`, `sec://...`) -- anything else (inline tokens, internal URLs, pasted customer data) now lands on disk in full. Proposed middle ground: keep this line truncated as before, and log the full command separately under a capped tag (e.g. `command_full`, mirroring how `command_output` is already logged) for `watch` to render.

src/tools/execute-prompt.ts
```diff
@@ -191,7 +191,7 @@ export async function executePrompt(input: ExecutePromptInput, extra?: any): Pro
     ? '⚠️ DEPRECATION: dangerously_skip_permissions is deprecated and ignored. Use update_member(unattended="dangerous") instead.\n\n'
     : '';
 
-  const scope = new LogScope('execute_prompt', `[${resolvedModel}] resume=${input.resume} timeout=${input.timeout_s ?? 300}s ${truncateForLog(maskSecrets(input.prompt))}`, agent);
+  const scope = new LogScope('execute_prompt', `[${resolvedModel}] resume=${input.resume} timeout=${input.timeout_s ?? 300}s ${singleLineForLog(maskSecrets(input.prompt))}`, agent);
```
kumaakh: Same log-retention concern as `execute-command.ts:135` -- `singleLineForLog` here means the full prompt text (masked-secrets-only) is now written uncapped to the fleet log instead of an 80-char snippet. Same proposed fix: truncate this line as before and add a separately-capped `prompt_full` tag for `watch`.

src/services/stall/log-path-resolver.ts
```diff
@@ -2,6 +2,17 @@ import type { LlmProvider } from '../../types.js';
...
 import { join, basename } from 'path';
 
+/**
+ * Claude Code stores each project's sessions under a single directory named
+ * after the project's absolute path, with EVERY non-alphanumeric character
+ * replaced by '-' (slashes, backslashes, colons, dots, underscores, spaces...).
+ * e.g. /home/ecs_user/vbv_nyk/app -> -home-ecs-user-vbv-nyk-app
+ * This must match Claude's own encoding exactly, or the transcript is not found.
+ */
+export function encodeClaudeProjectDir(workFolder: string): string {
```
kumaakh, line 12: Confirmed correct: this now replaces every non-alphanumeric char with `-`, matching Claude Code's actual on-disk directory-naming convention (the old `[/\:]`-only regex missed `_`/`.`, which was the bug). Good regression test added for the underscore case.

vbv-nyk:
Addressed both non-blocking follow-ups from the SSH review (`9aac77e`), plus my take on the log-retention question:

**1. `shq()` duplication (`watch.ts:384`)** -- dropped the local copy; it now imports the shared `escapeShellArg` from `src/utils/shell-escape.ts`, so there's a single escaping implementation to audit. The tail-command construction is factored into a small `buildTailCommand(startFlag, file)` helper that applies the escaping in one place.

**2. Missing adversarial test (`watch.ts:416`)** -- added `tests/watch-tail-command.test.ts`, which feeds hostile filenames through `buildTailCommand` and asserts they stay contained in a single quoted argument: `; rm -rf /`, a stray single quote (quote-break attempt), and `$(whoami)`/backtick command substitution. This locks in that the untrusted `ls -t` filename is escaped before it reaches `tail` over the SSH channel, so a future refactor can't silently drop it.

**3. Log retention / `singleLineForLog`** -- one thing worth flagging on the proposed `command_full`/`prompt_full` tag: it wouldn't actually make secrets any safer. The full (masked-only) text still lands on disk either way; moving it to a separate line in the same file doesn't reduce what's exposed. Its real benefits are narrower -- keeping the primary summary line short so nothing else parsing it changes, and re-adding a size cap so a large command/prompt can't bloat the log unbounded. Also worth noting the old `truncateForLog(.., 80)` wasn't reliable secret protection either -- a secret in the first 80 chars leaked before this change too; truncation only ever dropped whatever happened to sit past char 80. So if the goal is genuinely reducing secret exposure on disk, the lever is `maskSecrets` (more patterns) or not logging full text at all -- the separate tag is a growth/summary-line-stability improvement, not a security one.

Build clean; 56/56 across the watch + shell-escape suites pass (5 new).

kumaakh:
@vbv-nyk Re-reviewed at `9aac77e`. Build clean, **1638/1638 tests passing** (up from 1633 -- the 5 new `tests/watch-tail-command.test.ts` cases account for the delta).

## Addressed
**SSH escaping consolidation + adversarial test** -- fully resolved. `shq()` is gone; `buildTailCommand()` now uses the shared `escapeShellArg` from `src/utils/shell-escape.ts`, and `tests/watch-tail-command.test.ts` locks in the exact adversarial cases I asked for (`; rm -rf`, stray quote, `$(whoami)`/backticks all confirmed neutralized). This was the highest-risk item on the SSH surface and it's now fully covered. Nice work.

## Still open -- NOT addressed in this commit
The **log-retention** feedback (from my earlier inline comments on `src/tools/execute-command.ts:135` and `src/tools/execute-prompt.ts:194`) is untouched: both still call `singleLineForLog(maskSecrets(...))` with no cap, so full command/prompt text is still persisted to the fleet log rather than the old 80-char snippet. I want to flag this clearly since only one of the two open threads was addressed here -- wanted to make sure it wasn't dropped rather than deliberately deferred.

If the uncapped retention is an intentional tradeoff you're fine with (I can see the argument -- `watch` needs the full text to be useful), that's a reasonable call to make explicitly. If you'd rather address it, the proposed fix still stands: keep the primary log line truncated as before, and emit the full text on a separately-capped tag (e.g. `command_full`/`prompt_full`, same pattern as `command_output`) for `watch` to consume.

Either way, nothing else is blocking from my side -- just want an explicit decision on this one before merge.

vbv-nyk:
Implemented the log-retention fix. Summary of the approach:

**Fleet log persists only a truncated preview, length configurable.** `execute_command`/`execute_prompt` no longer write the full command/prompt -- they write `truncateForLog(maskSecrets(text), getLogPreviewChars())`. Preview length is a new `logging.previewChars` knob in config.json (default **256**; bad values warn and fall back). The log line goes back to being an identifier, not a full-text-on-disk record. Removed the now-dead `singleLineForLog`.

**`watch` still shows the full prompt -- from the transcript, single-sourced, no duplication.** The reason the log had gone uncapped was so `watch` could show the whole prompt. Turned out the transcript formatter was discarding the user prompt turn (it only rendered `tool_result`), so the log line was the only place the prompt surfaced. Now:
- The prompt renders from the transcript -- **400 chars by default, full under `-v`**.
- For any member whose transcript `watch` tails (local, or remote Claude over SSH), the fleet-log `execute_prompt` preview line is **dropped**, so the prompt shows exactly once. Members with no transcript (non-Claude, remote-without-transcript) keep the fleet-log preview.

Net: short line on disk, full prompt visible in `watch`, shown once.

Two limitations worth noting: (1) `execute_command` shell commands aren't in any transcript, so a truncated command line has no full-text fallback in `watch` -- commands are usually short and 256 covers most; (2) non-Claude providers' prompts show truncated-only in `watch` until transcript parsing lands (already deferred).

Verified: config knob honoured (80/256/512) against a real 768-char member prompt (-> 259 chars on disk); `watch` renders the prompt once against a live member with zero duplicate lines. Build clean; full suite 1659 passed / 5 skipped; new unit tests cover config parsing, the `-v` prompt cap, and the prompt-line suppression flag. Config lives at `~/.apra-fleet/data/config.json` (under `FLEET_DIR`).

src/cli/watch.ts
```diff
@@ -0,0 +1,539 @@
...
+  const collected: { f: Follower; ev: FormattedEvent }[] = [];
+  for (const line of lines) {
+    const entry = formatFleetLogLine(line, verbose);
+    if (!entry) continue;
+    const f = (entry.mem && byKey.get(entry.mem.toLowerCase())) || (entry.mid ? byKey.get(entry.mid) : undefined);
+    if (!f) continue;
+    // "Transcript owns the prompt": for members whose full transcript is tailed
+    // (local, or remote Claude), skip the fleet-log prompt preview so the prompt
+    // isn't shown twice. Members with no transcript keep the preview line.
+    if (entry.promptLine && (f.txDir || f.rtEnc)) continue;
```
kumaakh: This suppresses the fleet-log prompt preview whenever `f.rtEnc` is set -- but `rtEnc` just means "remote + claude provider" (set unconditionally at line 230), not "a transcript is actually being tailed right now." `ensureRemoteTail` fails soft on connect/exec errors (per its own doc comment) and just retries later, so during any period where the SSH channel is down, this condition still drops the prompt preview with nothing replacing it -- the prompt becomes invisible in `watch` for that stretch. Consider checking `f.rtStream` (the live channel) instead of `f.rtEnc` (static capability), or falling back to showing the preview line whenever `rtStream` is null.

kumaakh (commented):
@vbv-nyk Re-reviewed at `059ef6a`. Build clean, **1650/1650 tests passing** (up from 1638 -- new coverage in `tests/user-config.test.ts`, `tests/watch-fleet-log.test.ts`, `tests/watch-transcript-formatter.test.ts`).

## Log-retention feedback -- resolved well
This is a good fix, not just a bigger cap: `execute_command`/`execute_prompt` now write `truncateForLog(maskSecrets(text), getLogPreviewChars())` -- config-driven (`logging.previewChars`, default 256), validated with a warn-and-fallback on bad input. The fleet-log line goes back to being an identifier, and `watch` now sources the full prompt from the session transcript instead (400 chars by default, uncapped under `-v`). Dead `singleLineForLog` helper removed cleanly. This closes out both items I'd flagged as open.

## One new gap worth a look (non-blocking, but real)
See inline comment on `src/cli/watch.ts:359`. The fleet-log preview-suppression check uses `f.rtEnc` (a static capability flag: "this member is a remote Claude, so a transcript *should* be tailable") rather than whether the SSH tail is actually delivering content. `ensureRemoteTail` is explicitly documented as "fails soft" (network/auth errors just retry later) -- so if that channel is down, the fleet-log prompt preview is still suppressed even though nothing is replacing it. Net effect: a remote member with a struggling SSH tail can have its prompts invisible in `watch` with no indication why, which cuts against the feature's core goal (no more black-box members). Not a security issue, and probably a rare case, but worth either gating on `f.rtStream` (actual live channel) or falling back to the preview line when the channel isn't up.

## Minor
`logging.previewChars` isn't mentioned in `docs/features/watch.md`, `CHANGELOG.md`, or `README.md` -- worth a line so users know the knob exists.

Nothing else outstanding from my side. This is close -- just want your take on the rtEnc/rtStream distinction before I'd call it fully done.

src/cli/watch.ts
vbv-nyk-apra: Fixed at d865d27: the suppression condition now checks `f.rtStream` (the live SSH tail channel) instead of `f.rtEnc` (static remote+claude capability). If the channel is down or still connecting, `rtStream` is null and the fleet-log prompt preview falls back to showing, so the prompt is no longer dropped during that window. Build clean, full suite passing (2 pre-existing unrelated failures in `time-utils.test.ts` only).

kumaakh (approved):
@vbv-nyk Verified `d865d27` -- confirmed by reading the diff (not just the commit message): the suppression condition now checks `f.rtStream` (the live SSH channel) instead of `f.rtEnc` (static capability), exactly as suggested. When the channel is down or still connecting, `rtStream` is null and the fleet-log preview line falls back to showing, so the prompt is no longer dropped during that window.

Build clean, full suite passing: **1650/1650** (100 files, 14 skipped -- all pre-existing/environment-related skips, no failures).

Running tally of everything raised across this review:
- SSH command-injection safety (remote tail filename escaping) -- confirmed safe, then consolidated to the shared `escapeShellArg` helper with adversarial-input regression tests added.
- Log-retention (full command/prompt text persisted uncapped) -- fixed with a config-driven, validated preview cap + transcript-sourced full-text display in `watch`.
- Fleet-log prompt-preview dropped during SSH channel outages -- fixed, verified above.
- Session-dir path-encoding bug -- fixed and regression-tested.

Nothing outstanding from my side. **Approving.**

(Minor, non-blocking, no action needed before merge: `logging.previewChars` still isn't mentioned in `docs/features/watch.md`/`CHANGELOG.md`/`README.md` -- fine to pick up in a follow-up doc pass.)

PR #326, opened 2026-07-08
fix: MCP execute-command cwd alias + agy prompt/exit fixes

kumaakh:
## Summary
- `execute-command.ts`: accept a `cwd` alias for `run_from` so a strict-schema `cwd` field doesn't crash the MCP tool call.
- `windows.ts` / `agy.ts`: build an absolute path for the task prompt file instead of relying on "current directory" text, fixing prompt-resolution issues when the working directory isn't what the agent expects.
- `strategy.ts`: switch the child-process cleanup listener from `close` to `exit`.
- `agy.ts`: normalize a quoted `"agy"` filePath before use.
- `AGENTS.md`: ASCII dash fix, plus a new "Permissions (pre-grant rule)" section (added by the `agy-auto-sprint` commit — see review comments).

## Test plan
- [x] `npm test` — 100 files / 1650 tests passed, 14 skipped, 0 failed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,4 +1,4 @@
-# Apra Fleet — Agent Context
+# Apra Fleet - Agent Context
 
 Read `README.md` in this repo for the full tool reference, installation, member registration, multi-provider setup, git authentication, PM skill commands, and troubleshooting.
 
@@ -18,3 +18,15 @@ node dist/index.js install     # Dev-mode install
 - Never push to `main` directly; open a PR
 - See [Architecture](docs/architecture.md) for internal structure
 - ASCII only: never write non-ASCII characters to any file. Use `-` for dashes, `->` for arrows, `[OK]` for checkmarks, etc.
+
+## Permissions (pre-grant rule)
+
+At the start of any task that involves multiple file writes:
+
+1. Call `ask_permission(Action='write_file', Target=<narrowest directory covering all writes>)` ONCE before any file is written.
+2. Never call `ask_permission` again for files inside an already-granted directory in the same session.
+3. For pm-lite sprints: grant the worktree root immediately after `git worktree add`, before dispatching any subagent.
+4. For single-repo work: grant the repo root at task start.
+5. Subagents inherit the grant - the orchestrator requests it once on their behalf before spawning them.
+6. Never call `ask_permission` for an individual file when its parent directory is already granted.
+
--- a/src/os/windows.ts
+++ b/src/os/windows.ts
@@ -100,7 +100,8 @@ export class WindowsCommands implements OsCommands {
   buildAgentPromptCommand(provider: ProviderAdapter, opts: PromptOptions): string {
     const { folder, promptFile, sessionId, resuming, unattended, model, maxTurns, inv } = opts;
     const escapedFolder = escapeWindowsArg(folder);
-    let instruction = `Your task is described in ${promptFile} in the current directory. Read that file first, then execute the task.`;
+    const absPath = folder.replace(/\\/g, '/') + '/' + promptFile;
+    let instruction = `Your task is described in the file at ${absPath}. Read that absolute path first, then execute the task.`;
     if (inv) {
       instruction = `[${inv}] ${instruction}`;
     }
--- a/src/providers/agy.ts
+++ b/src/providers/agy.ts
@@ -53,7 +53,8 @@ export class AgyProvider implements ProviderAdapter {
   buildPromptCommand(opts: PromptOptions): string {
     const { folder, promptFile, sessionId, resuming, unattended, inv, model, tier: inputTier } = opts;
     const escapedFolder = escapeDoubleQuoted(folder);
-    let instruction = `Your task is described in ${promptFile} in the current directory. Read that file first, then execute the task.`;
+    const absPath = folder.replace(/\\/g, '/') + '/' + promptFile;
+    let instruction = `Your task is described in the file at ${absPath}. Read that absolute path first, then execute the task.`;
     if (inv) {
       instruction = `[${inv}] ${instruction}`;
     }
@@ -221,6 +222,7 @@ export class AgyProvider implements ProviderAdapter {
   }
 
   wrapWindowsPrompt(setupCmd: string, filePath: string, argList: string, sessionId?: string, model?: string, tier?: 'cheap' | 'standard' | 'premium'): string {
+    if (filePath === '"agy"') filePath = 'agy';
     // Write per-workspace model override before launching agy (mirrors buildPromptCommand).
     const resolvedTier = tier ?? this.resolveTierFromModel(model);
     const displayModel = getModelOverride('agy', resolvedTier) ?? AGY_MODEL_FOR_TIER[resolvedTier];
--- a/src/services/strategy.ts
+++ b/src/services/strategy.ts
@@ -155,7 +155,7 @@ class LocalStrategy implements AgentStrategy {
         }
       });
 
-      child.on('close', (code) => {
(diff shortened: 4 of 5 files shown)


AGENTS.md
```diff
@@ -18,3 +18,15 @@ node dist/index.js install     # Dev-mode install
 - Never push to `main` directly; open a PR
 - See [Architecture](docs/architecture.md) for internal structure
 - ASCII only: never write non-ASCII characters to any file. Use `-` for dashes, `->` for arrows, `[OK]` for checkmarks, etc.
+
+## Permissions (pre-grant rule)
+
+At the start of any task that involves multiple file writes:
```
kumaakh, line 24: **BLOCKING:** This whole "Permissions (pre-grant rule)" section is unrelated to the stated purpose of this commit (a narrow `cwd`-alias bug fix) and has no rationale in the commit message, PR description, or `docs/`. I grepped `src/` and `tests/` — `ask_permission` is not implemented anywhere in this repo; this is prescriptive prose aimed at an external agent runtime. Substantively it tells agents to request one whole-directory write grant and never re-confirm for any file under it, and to let subagents inherit that grant automatically without re-asking. That widens the blast radius of a compromised or hallucinating (sub)agent from a single file to an entire directory tree, with nothing in code enforcing scope or expiry. Please split this into its own PR with an explicit rationale, or remove it — this should not be merged silently inside a bug-fix commit.

src/services/strategy.ts
```diff
@@ -155,7 +155,7 @@ class LocalStrategy implements AgentStrategy {
         }
       });
 
-      child.on('close', (code) => {
+      child.on('exit', (code) => {
```
kumaakh, line 158: **MAJOR:** Switching from `child.on('close', ...)` to `child.on('exit', ...)` changes flush semantics. `exit` fires when the child process terminates, before stdout/stderr pipes are guaranteed to have flushed their final `data` events, whereas `close` waits for the stdio streams to fully close. This handler immediately calls `stdoutSpillStream.end()`/`stderrSpillStream.end()` and resolves `{stdout, stderr, code}` — so this risks truncating the last chunk(s) of captured output if data arrives after the child exits but before the pipe drains. This may well be an intentional fix for a real Windows issue (`close` never firing because a detached grandchild still holds the stdio handle open, hanging the strategy), but there's no test covering either behavior and no explanation in the commit message. Please share the repro/rationale, and consider adding a regression test for whichever failure mode this addresses.

src/providers/agy.ts
```diff
@@ -221,6 +222,7 @@ export class AgyProvider implements ProviderAdapter {
   }
 
   wrapWindowsPrompt(setupCmd: string, filePath: string, argList: string, sessionId?: string, model?: string, tier?: 'cheap' | 'standard' | 'premium'): string {
+    if (filePath === '"agy"') filePath = 'agy';
```
kumaakh, line 225: **MAJOR (dead code):** Traced the only call site — `windows.ts:113` sets `filePath = provider.cliCommand('').trim()`, and `AgyProvider.cliCommand(args)` returns ```agy ${args}```, which for `args=''` trims to the *unquoted* string `'agy'`, never `'"agy"'`. This condition can never be true through the actual code path in this repo, so it's either leftover debug code targeting a caller/scenario not present here, or the real bug is elsewhere and this doesn't fix it. Recommend removing it or root-causing the actual quoting issue it was meant to address.

src/tools/execute-command.ts
```diff
@@ -122,7 +122,8 @@ function redactOutput(output: string, credentials: ResolvedCredential[]): string
   return redacted;
 }
 
-export async function executeCommand(input: ExecuteCommandInput, extra?: any): Promise<string> {
+export async function executeCommand(input: ExecuteCommandInput & { cwd?: string }, extra?: any): Promise<string> {
+  if (input.cwd && !input.run_from) input.run_from = input.cwd;
```
kumaakh, line 126: **Minor:** `ExecuteCommandInput`'s Zod schema has no `cwd` field — it's only introduced via the inline `&amp; { cwd?: string }` intersection type, so this bypasses schema validation entirely and mutates the input object in place. Harmless today (grepped — nothing else in `src/`/`tests/` treats `cwd` as a semantic field, so no shadowing risk), but it's an undocumented, unvalidated back door into the schema. Consider adding `cwd` to the Zod schema itself so validation and the alias stay in sync.

src/os/windows.ts
```diff
@@ -100,7 +100,8 @@ export class WindowsCommands implements OsCommands {
   buildAgentPromptCommand(provider: ProviderAdapter, opts: PromptOptions): string {
     const { folder, promptFile, sessionId, resuming, unattended, model, maxTurns, inv } = opts;
     const escapedFolder = escapeWindowsArg(folder);
-    let instruction = `Your task is described in ${promptFile} in the current directory. Read that file first, then execute the task.`;
+    const absPath = folder.replace(/\\/g, '/') + '/' + promptFile;
```
kumaakh, line 103: **Minor:** `absPath` is built from raw `folder` with no escaping (`folder.replace(/\\/g, '/') + '/' + promptFile`), even though `escapedFolder` (via `escapeWindowsArg`) is computed on the line above and available. Since the whole instruction string ends up as one quoted `-p` argument, spaces in `folder` are fine, but a literal `"` in `folder` would break out of that quoted argument. Low likelihood on Windows paths, but worth using the already-escaped value for consistency, and no test exercises a folder with special characters.

kumaakh:
## Automated review summary

Tests: `npm test` — 100 files / 1650 tests passed, 0 failed, 14 skipped. No regressions detected in the suite.

Left 5 inline review comments. **Not marking this approved** — GitHub blocks self-review/self-approval since I'm also the PR author, but the verdict is **changes requested**, driven by one blocking and two major items:

- **BLOCKING** ([AGENTS.md](https://github.com/Apra-Labs/apra-fleet/pull/326#discussion_r3547423450)): the new "Permissions (pre-grant rule)" section is unrelated to this commit's stated purpose (a narrow `cwd`-alias fix), documents a tool (`ask_permission`) that doesn't exist anywhere in this codebase, and instructs agents to grant themselves whole-directory write access once and never re-confirm — including for subagents inheriting the grant automatically. That's a real privilege-broadening change with no code enforcing scope/expiry, bundled silently into a bug-fix commit. Please split it into its own PR with explicit rationale, or drop it.
- **MAJOR** ([strategy.ts:158](https://github.com/Apra-Labs/apra-fleet/pull/326#discussion_r3547423514)): `close` -> `exit` on the child-process listener risks truncating spill-stream output, since `exit` doesn't guarantee stdio has finished flushing before the handler ends the streams. Possibly a legitimate Windows fix (hung `close` because a detached grandchild still holds the stdio handle open) but needs the rationale/repro and ideally a regression test.
- **MAJOR** ([agy.ts:225](https://github.com/Apra-Labs/apra-fleet/pull/326#discussion_r3547423576)): the `filePath === '"agy"'` check is dead code — traced the only call site (`windows.ts:113` -> `cliCommand('').trim()`) and it always yields the unquoted string `'agy'`, so this branch can never trigger through the real code path. Remove it or root-cause the actual quoting issue it was meant to fix.

Two minor, non-blocking nits: [execute-command.ts:126](https://github.com/Apra-Labs/apra-fleet/pull/326#discussion_r3547423629) (`cwd` bypasses the Zod schema) and [windows.ts:103](https://github.com/Apra-Labs/apra-fleet/pull/326#discussion_r3547423693) (unescaped `absPath`).

Everything else in the diff (the `cwd` alias mechanism, the absolute-path prompt fix) looks correct and low-risk.

PR #364, opened 2026-07-29
fix(bundle): make the packaged fleet-sprint binary actually run

joiskash:
The `fleet-sprint` binary declared in this package's `bin` map
(`dist/fleet-sprint.mjs`, installed onto PATH by
`npm install -g @apralabs/apra-fleet`) cannot start on `main` today. Two
independent breakages, one masking the other:

1. **Dies on startup.** The bundle is `format: 'esm'` but pulls in CJS deps --
   undici, via `@apralabs/apra-fleet-client`'s HTTP transport. esbuild rewrites
   CJS `require(...)` into a `__require` shim whose fallback throws, and undici
   hits it at runtime from `lib/dispatcher/client.js`:
   `Dynamic require of "node:assert" is not supported`. Every invocation of that
   binary died before parsing a flag.

   Fixed with a `createRequire` banner. The shim's own first branch is
   `typeof require !== "undefined" ? require : <throwing fallback>`, so putting
   a genuine `require` in module scope makes it take the working branch. This
   is the canonical esbuild ESM/CJS interop fix, not an undici-specific patch --
   any future CJS dep doing a dynamic require is covered. esbuild keeps the
   entry's shebang on line 1 ahead of the banner, so the bin stays executable.

2. **Ships an incomplete runner.** `runner.js` is deliberately never bundled
   (`engine.executeFile()` reads it from disk as text), so its relative
   `./x.mjs` imports must resolve against real files next to
   `dist/fleet-sprint-runner.mjs`. `bundle-se.mjs` copied a hardcoded three;
   `conflict-ladder.mjs` and `sprint-lock.mjs` became runner imports later and
   never shipped, so the binary died with `ERR_MODULE_NOT_FOUND` for
   `dist/conflict-ladder.mjs` on the first real sprint.

   Fixed by scanning the directory instead of naming files, so a new sibling
   ships the moment it exists. This is already earning its keep: #362 added
   `full-db-fetch-guard.mjs`, which the hardcoded list would also have dropped.

## Scope: which entry point this affects

`packages/apra-fleet-se/fleet-sprint/docs/README.md` documents two ways to run
a sprint, and **neither goes through this bundle**:

1. `apra-fleet workflow fleet-sprint` -- resolves from
   `~/.apra-fleet/workflows/fleet-sprint`, which `install.ts` populates from
   `collectPackageTree(packages/apra-fleet-se)`, i.e. the SE *source tree*.
2. `node packages/apra-fleet-se/bin/cli.mjs` -- the unbundled dev entry point.

This PR fixes the third, undocumented-but-shipped path: the `fleet-sprint` bin
itself, which `bundle-se.mjs`'s own header states is there so
`npm install -g @apralabs/apra-fleet` "also gets a working `fleet-sprint`
binary". It currently does not. So the blast radius is `npm i -g` users
invoking `fleet-sprint` directly, not the documented workflows.

## Why nothing caught this

No test had ever executed the packaged artifact.
`cli-server-resolution.test.mjs` only writes a fake `// bundled cli`
placeholder, so both breakages sat in the shipped binary while the documented
paths (which bypass it) kept working. Bug 1 also masked bug 2 -- the binary
never got far enough to load the runner, so the missing modules could not
surface; fixing the first exposed the second on the very next launch.

`tests/fleet-sprint-bundle-smoke.test.ts` closes that gap: it spawns the built
bundle and requires it to reach flag parsing, asserts the shebang stays on line
1, and walks the runner's relative `.mjs` imports transitively so a missing
sibling fails the build rather than a user's sprint. The `--help` smoke alone
cannot catch bug 2 -- the runner loads at sprint-run time, long after `--help`
exits 0. Skipped (not failed) when `dist/fleet-sprint.mjs` is absent, since it
is built by `build:se`/`prepublishOnly`, not a plain `npm run build`.

Both fixes were verified as real regression tests: reverting each makes its
test fail with the exact original error.

## Also included

`tests/install-dev-manifest.test.ts` is a guard, not a fix -- the paths are
already correct on `main`. `buildDevManifest()`'s workflow-subsystem inputs are
hardcoded filesystem paths, and `hasWorkflowSubsystemAssets()` requires all
three sections, so one stale path silently disables the whole workflow install:
no `~/.apra-fleet/workflows/`, no non-zero exit, and an install that still
prints success and records `workflowsMode:"all"` -- which breaks documented
invocation form 1 above. That was observed for real on a pre-`bd01665` tree
where `agentSchemasDir` still pointed at the retired `vendor/apra-pm`
submodule. Nothing stops the next directory move from repeating it.

## Testing

- Root: 2740 passed / 10 skipped. One unrelated pre-existing flake --
  `tests/hub-service/relay-queue.test.ts` times out at 20s under parallel load
  and passes on retry in isolation; this branch touches no hub-service file.
- `packages/apra-fleet-se`: 1065 passed, 0 failed.

## Not included

An earlier revision of this branch also carried an `Ensure Sprint Branch` fix
for a local-only branch being force-reset to base (real data loss: it destroyed
three unpushed commits). `bd01665`'s `decideEnsureBranchAction()` supersedes it
and is strictly better -- it also handles the local-ahead-of-origin ancestry
case that fix left open -- so it was dropped rather than merged.

### DIFF
--- a/scripts/bundle-se.mjs
+++ b/scripts/bundle-se.mjs
@@ -149,6 +149,25 @@ await build({
   minify: false, // keep readable for debugging
   // cli.mjs's own shebang (#!/usr/bin/env node) is preserved by esbuild
   // automatically when it is the first line of the entry point.
+  //
+  // ESM output + a bundled CJS dependency needs a real `require` in scope.
+  // esbuild rewrites CJS `require(...)` calls into its own `__require` shim,
+  // and that shim's fallback THROWS -- `Dynamic require of "node:assert" is
+  // not supported` -- for any call it could not resolve statically. undici
+  // (pulled in transitively by @apralabs/apra-fleet-client's HTTP transport)
+  // does exactly that at runtime from lib/dispatcher/client.js, so every
+  // invocation of the bundled binary died on startup before parsing a single
+  // flag. The shim's first branch is `typeof require !== "undefined" ?
+  // require : <throwing fallback>`, so defining a genuine createRequire-backed
+  // `require` at module scope makes it take the working branch instead. This
+  // is the canonical esbuild ESM/CJS interop fix, not a workaround for undici
+  // specifically: any future CJS dep doing a dynamic require is covered.
+  banner: {
+    js: [
+      "import { createRequire as __apraCreateRequire } from 'node:module';",
+      'const require = __apraCreateRequire(import.meta.url);',
+    ].join('\n'),
+  },
   metafile: true,
 });
 
--- a/scripts/bundle-se.mjs
+++ b/scripts/bundle-se.mjs
@@ -79,7 +79,7 @@
 // the two @apralabs/* workspace packages below (apra-fleet-7pm.12).
 
 import { build } from 'esbuild';
-import { copyFileSync, cpSync, existsSync, mkdirSync, readFileSync, rmSync } from 'node:fs';
+import { copyFileSync, cpSync, existsSync, mkdirSync, readdirSync, readFileSync, rmSync } from 'node:fs';
 import { join, dirname } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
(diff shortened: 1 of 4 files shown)


joiskash:
Partly superseded, partly obsolete — see #305.

`2cc52680` (install dev-manifest guard) **was cherry-picked into #305**. It is a real regression guard for a failure observed live — a stale `vendor/apra-pm` path silently disabling the entire workflow install while the installer still printed success — and it still applies.

The two bundle commits are **obsolete**. `533fee23` (ESM `require` for CJS deps) and `db7367fa` (ship every runner sibling) both patch `scripts/bundle-se.mjs`, which `main` retired in apra-fleet-fyc.2/kuh.5. Per `docs/npm-packaging.md` on `main`:

> **the fleet-sprint engine now ships as source, not a bundle.** `scripts/bundle-se.mjs` and the `build:se` script have been retired

With no esbuild bundle there is no ESM/CJS require problem and no hand-maintained sibling subset, so both bugs are moot by construction. `main`'s own `docs/test-inventory.md` already records the matching smoke test as dead: *"the guard at `:35` can never be satisfied."*

PR #391, opened 2026-08-10
feat(dispatch): pre-flight readiness check before task dispatch

omkarsm:
## Summary

- Adds a lightweight pre-dispatch readiness check (`src/services/preflight-check.ts`) that verifies member connectivity and LLM auth BEFORE the expensive prompt/command dispatch
- Catches expired OAuth tokens, missing credentials, and offline members in <1s instead of burning a full dispatch round trip
- Integrates into both `execute_prompt` (full check: connectivity + auth) and `execute_command` (connectivity only)
- Invalidates the preflight cache in `provision-auth.ts`, `update_member`, `remove_member`, cloud `lifecycle.ts`, and `setup-ssh-key.ts` after state changes
- Skips preflight for interactive/channelCapable sessions (they have their own liveness probes)

## Motivation

Auth failures were wasting full dispatch round trips (30-60s) before being classified. The preflight check catches this in <1s.

## Design

The check is deliberately lightweight (no real LLM API call):
1. **Connectivity**: SSH `testConnection()` (reuses pooled connection)
2. **Credential presence**: file-exists check for OAuth files or env-var check for API keys
3. **OAuth freshness**: reads the remote credential JSON and validates `expiresAt` (Claude provider; other providers log a warning and pass)

Results are cached for 60s (`CACHE_TTL_MS`), scoped by auth level (`conn` vs `full`) to prevent cache poisoning across `skipAuth` boundaries. Cache has a 500-entry size cap with oldest-entry eviction.

## Review feedback addressed (round 1)

All 8 findings from @kumaakh's review have been fixed:

1. **F1** `execute-command.ts`: Returns `structuredContent` with `isError: true, reason: 'preflight_failed'` (was bare string)
2. **F2** `preflight-check.ts`: Cache keys scoped by auth level (`${id}:conn` / `${id}:full`)
3. **F3** `execute-prompt.ts`: Interactive `channelCapable` sessions skip preflight entirely
4. **F4** `preflight-check.ts`: Non-Claude providers get a log warning for OAuth freshness
5. **F5** `preflight-check.ts`: Uses `cmds.readRemoteJson()` instead of hand-rolled file reads
6. **F6** `preflight-check.ts`: `clearTimeout()` in both success and catch paths of connectivity race
7. **F7** `update-member.ts` + `remove-member.ts`: `invalidatePreflightCache()` on identity-affecting changes
8. **F8** `preflight-check.ts`: OAuth + API key checks run in parallel via `Promise.all`

## Review feedback addressed (round 2)

8 additional findings from @kumaakh's second review:

1. **R2-F1**: Added `invalidatePreflightCache` to `cloud/lifecycle.ts` (IP rotation/reattach) and `setup-ssh-key.ts` (auth type change)
2. **R2-F2**: Handle `CLAUDE_CODE_OAUTH_TOKEN` env var via `provider.authEnvVarForToken()` probes; filter empty/invalid env var names to prevent synchronous throw for OpenCode-like providers
3. **R2-F3**: Re-query `sessionRegistry` after preflight await to avoid stale session snapshots
4. **R2-F4**: Check `apiKeyResult` before early-returning on expired OAuth (API key is a valid fallback)
5. **R2-F5**: Use distinct preflight reason codes (`preflight_offline`, `preflight_auth_missing`, `preflight_auth_expired`) for fleet-sprint error classification
6. **R2-F6**: Added `input.username` to `identityChanged` gate in `update-member.ts`
7. **R2-F7**: Added `.catch(() => {})` on `testConnection` after Promise.race timeout to prevent unhandled rejection
8. **R2-F8**: Added cache size cap (CACHE_MAX_SIZE=500) with oldest-entry eviction

## Test plan

- [x] Global test mock in `tests/setup.ts` prevents preflight from breaking existing test suites
- [x] `tests/preflight-check.test.ts` uses `vi.unmock` to test the real implementation
- [x] Regression test for OpenCode-like provider with empty authEnvVar
- [x] `tests/update-member.test.ts` -- invalidatePreflightCache assertions for identity changes
- [x] `tests/cloud-lifecycle-unit.test.ts` -- invalidatePreflightCache for cloud lifecycle
- [x] `tests/setup-ssh-key.test.ts` -- setup-ssh-key coverage including cache invalidation
- [x] `tests/execute-prompt.test.ts` -- preflight_* reason code mapping
- [x] `packages/apra-fleet-se/test/preflight-reason-classification.test.mjs` -- errors.mjs truth table
- [x] CI green on all 3 platforms


kumaakh:
This is a good step forward. In hindsight I added many of these simialr checks in fleet-sprint while I should have focussed on improving the core. 

I understand why execute_prompt must have a pre-flight check (LLM creds may be missing/expired) 
I understand why execute_command needs a preflight check because remote member may go offline. 
I realize that the implementation is common : both call the same pre-flight check service.  

have you considered the case for "no LLM" members @omkarsm ?

kumaakh:
@omkarsm CI is red on this PR -- root-caused it (background investigation), posting the findings here.

## Root cause: `preflightCheck()` breaks every existing dispatch-test mock

**148 tests failed on `ubuntu-latest`** (`windows-latest`/`macos-latest` show "fail" too, but their logs end in `The operation was canceled` mid-suite -- they were killed by matrix fail-fast once ubuntu failed; not independent failures). main's last CI run (31402503975, sha 4ca0ff01) is green, so all 148 are new on this branch.

This PR adds a `preflightCheck(agent)` call before every non-`local`-member dispatch, in:
- `src/tools/execute-prompt.ts` (~line 576-594)
- `src/tools/execute-command.ts` (~line 148-155)

`tests/test-helpers.ts:33` makes test agents `agentType: 'remote'`, so every *existing* execute_prompt/execute_command test now runs through the new preflight path.

`preflightCheck` (`src/services/preflight-check.ts`, step 1) does:
```ts
const conn = await Promise.race([strategy.testConnection(), ...timeout]);
if (!conn.ok) { ... }
```
Existing suites mock `getStrategy` with a bare `testConnection: vi.fn()` (e.g. `tests/execute-command.test.ts:13`) or an un-stubbed `mockTestConnection` (`tests/tool-provider.test.ts:23-31`), so `testConnection()` resolves `undefined`, `conn.ok` throws a `TypeError`, and the catch turns that into `code: 'offline'`. Every dispatch then short-circuits before it ever reaches the scenario the test was written to exercise:
- `executeCommand` -> `[FAIL] Pre-dispatch check failed for "X": Member "X" is unreachable: Cannot read properties of undefined (reading 'ok')`
- `executePrompt` -> `structuredContent: { isError: true, reason: 'dispatch_failed' }`

That explains every distinct error signature in the log: `expected 'dispatch_failed' to be 'empty_response'/'max_turns_exhausted'/'session_not_found'/...`, `expected "vi.fn()" to be called N times, but got 0 times`, `Cannot read properties of undefined (reading '0')` (from `mockExecCommand.mock.calls[0]` when execCommand was never reached), and `expected '[FAIL] Pre-dispatch check failed for ...' to contain '<expected output>'`.

Affected files: `execute-prompt` (.test, -interactive, -provisioning, -resume-semantics, -orphan-recovery, -substitution, -phantom-connectback, -idb-busy-lock), `execute-command`, `tool-provider`, `model-tiers`, `unattended-mode`, `cloud-lifecycle`, `dead-session-detection-contract`, `member-reservation-e2e`, `stall-any-entry-activity-and-kill`, `stop-prompt`, `integration/session-lifecycle`. (`tests/preflight-check.test.ts`, this PR's own new suite, passes.)

## Fix needed -- two parts, the second is easy to miss

1. **Connectivity mocks.** Guard `preflightCheck` (`if (!conn?.ok)`) *and* have every affected suite stub `testConnection` to resolve `{ ok: true, latencyMs: 1 }`. Defensive-only isn't enough -- a resolved-`undefined` still makes `ok` falsy, so dispatch still fails closed.
2. **Call-count/index shift.** Even with `testConnection` stubbed green, `preflightCheck` steps 2-3 issue up to three extra `strategy.execCommand` calls (`cmds.credentialFileCheck`, the `cat`/`Get-Content` read, `cmds.apiKeyCheck`). Those consume the shared `mockExecCommand` queue, so `expect(mockExecCommand).toHaveBeenCalledTimes(3)`-style assertions and all `mock.calls[N]` index assertions will still break across ~15 files. The 60s `preflightCache` doesn't help here since `makeTestAgent` mints a fresh random id per test. Options: thread `{ skipAuth: true }` through `executePrompt` as well as `executeCommand`, route the auth probe off `strategy.execCommand` entirely, or add a test-env opt-out.

## Two more defects noticed while reading (not CI-visible, but worth a bead)

- **Cache poisoning across the `skipAuth` boundary** (`preflight-check.ts`, step 2): `execute_command` calls with `skipAuth: true` still write a success entry to `preflightCache.set(agent.id, ...)`. A subsequent `execute_prompt` on the same member within the 60s window gets a cache hit and skips the auth check entirely -- exactly the check this PR exists to add. The cache key should encode whether auth was actually validated, not just connectivity.
- **Timer leak**: the `Promise.race` timeout's `setTimeout` is never cleared when `testConnection` wins the race, leaving a pending 10s timer per preflight call.

Key files: `src/services/preflight-check.ts`, `src/tools/execute-prompt.ts`, `src/tools/execute-command.ts`, `tests/test-helpers.ts`, `tests/tool-provider.test.ts`, `tests/execute-command.test.ts`.

(Separately, `/code-review` is running on this PR too -- I'll post/point to those findings once that lands.)

kumaakh:
@omkarsm follow-up: the separate \`/code-review\` pass finished and independently confirmed the CI root-cause findings above, plus 3 more worth fixing alongside:

1. **\`src/services/preflight-check.ts:133\`** -- the OAuth freshness check only understands Claude's JSON credential shape; it silently no-ops for Gemini members, so preflight gives no real signal there.
2. **\`src/tools/execute-prompt.ts:585\`** -- the preflight gate sits ahead of the interactive-routing path, which was specifically designed not to need SSH/relay connectivity. Worth checking whether interactive dispatch should skip preflight entirely.
3. **\`src/services/preflight-check.ts:141\`** -- reimplements \`readRemoteJson()\`/\`readTextFile()\` locally with weaker, inconsistent escaping instead of reusing the existing helpers.

(The other 3 findings from that pass -- \`execute-command.ts:153\` returning a bare string with no \`structuredContent\`/\`isError\` so failures get silently treated as success by \`command()\`, the cache-poisoning issue, and the timer leak -- match what's already noted above.)

kumaakh:
@omkarsm final consolidated \`/code-review\` report for this PR (supersedes my last two comments -- same underlying issues, now with precise line numbers and full failure scenarios), ranked most severe first:

1. **\`src/tools/execute-command.ts:153\`** -- on preflight failure, \`executeCommand\` returns a bare \`'[FAIL] ...'\` string with no \`structuredContent\`, so \`apra-fleet-workflow\`'s \`command()\` (packages/apra-fleet-workflow/src/workflow/index.mjs:1249-1276) only throws on specific text prefixes or \`structuredContent.exitCode !== 0\` -- neither applies here, so \`exitCode\` resolves to \`null\` and the failure is silently treated as normal stdout. Every workflow caller (fleet-sprint, apra-pm scripts) dispatching to a member with expired/missing auth gets no \`CommandError\`. Also violates \`packages/apra-fleet-client/src/client/errors.mjs\`'s own documented convention (2026-07-19): new failure kinds downstream code needs to branch on must get a typed error class or reason code, never a bare string to sniff.
2. **\`src/services/preflight-check.ts:84\`** -- the 60s pass-cache doesn't distinguish a connectivity-only (\`skipAuth\`) pass from a full auth-checked pass. \`execute-command.ts\` calls \`preflightCheck(agent, {skipAuth:true})\`, caches \`{passedAt}\` keyed only by \`agent.id\`; within 60s an \`execute-prompt.ts\` call hits that cache and skips OAuth-expiry/credential checks entirely. A member with expired/missing auth that recently ran \`execute_command\` passes preflight for \`execute_prompt\` with dead credentials.
3. **\`src/tools/execute-prompt.ts:585\`** -- preflight (SSH/relay connectivity + auth) now unconditionally gates \`execute_prompt\` *before* the interactive-routing branch, which dispatches via a live MCP push channel instead of SSH. A member reachable only via a live \`channelCapable\` session but whose SSH \`testConnection()\` currently fails now gets rejected as \`offline\` before interactive routing is ever consulted -- contradicting the new code's own comment that "interactive sessions have their own liveness probes."
4. **\`src/services/preflight-check.ts:152\`** -- OAuth freshness validation only understands Claude's credential JSON shape (\`parsed.claudeAiOauth\`), so it silently never fires for other providers. A Gemini/Codex/Copilot/AGY member with an expired, non-refreshable token passes preflight on file-existence alone.
5. **\`src/services/preflight-check.ts:148\`** -- the Windows OAuth-file read is hand-rolled (\`powershell -Command\` with manual quote-escaping) and the Linux read has no escaping, duplicating \`readTextFile()\`/\`escapeWindowsArg\` in \`src/os/windows.ts\`/\`src/os/linux.ts\`. Not exploitable today (paths are static provider literals), but this is a second, differently-escaped path that won't inherit future escaping fixes made to the shared helpers.
6. **\`src/services/preflight-check.ts:97\`** -- the \`Promise.race\` connectivity-timeout \`setTimeout\` is never cleared when \`testConnection()\` wins. On every warm-connection dispatch (the common case) a live 10s timer keeps running post-race, unlike \`ssh.ts\`/\`strategy.ts\` which \`clearTimeout\`/\`.unref()\`.
7. **\`src/services/preflight-check.ts:56\`** -- \`invalidatePreflightCache\` is only wired into \`provision-auth.ts\`; nothing invalidates it on \`update_member\` (host/port/credential edits) or member removal. Rotating a member's SSH host/port within the 60s cache window leaves the stale \`ok=true\` entry masking the change.
8. **\`src/services/preflight-check.ts:141\`** -- the OAuth file-existence check, file-content read, and API-key env-var check run as three independent sequential SSH round trips instead of parallel/combined, partly offsetting the PR's own dispatch-latency goal for providers with both \`oauthCredentialFiles\` and \`authEnvVar\` configured.

Key files: \`src/services/preflight-check.ts\` (new), \`src/tools/execute-command.ts\`, \`src/tools/execute-prompt.ts\`, \`src/tools/provision-auth.ts\`, \`packages/apra-fleet-workflow/src/workflow/index.mjs\`, \`packages/apra-fleet-client/src/client/errors.mjs\`.

omkarsm:
@kumaakh All 8 findings addressed in commits `900d7cb9..332ede5a`:

| # | Finding | Fix commit |
|---|---------|-----------|
| F1 | executeCommand bare string | `01b21c6f` -- returns structuredContent with reason: 'preflight_failed' |
| F2 | Cache poisoning across skipAuth | `f6dcfcac` -- cache keys scoped by auth level (conn/full) |
| F3 | Preflight gates interactive | `f6dcfcac` -- channelCapable sessions skip preflight |
| F4 | Non-Claude OAuth no-op | `2457326e` -- logs warning for non-Claude providers |
| F5 | Hand-rolled file reads | `2457326e` -- uses cmds.readRemoteJson() from src/os/ |
| F6 | Timer leak in Promise.race | `332ede5a` -- clearTimeout in success + catch paths |
| F7 | Cache not invalidated on update/remove | `332ede5a` -- wired into update_member + remove_member |
| F8 | Sequential SSH round trips | `332ede5a` -- Promise.all for OAuth + API key checks |

CI is green on all 3 platforms. Ready for re-review.

kumaakh (commented):
Automated re-review (fresh pass against the current/reworked diff). 8 findings below, all around the new preflight-check auth/connectivity feature.

src/services/preflight-check.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  // Process API key result
+  if (apiKeyResult) {
+    apiKeyPresent = apiKeyResult.stdout.trim().length > 5;
+  }
+
+  // Also check stored encrypted env vars (provision-auth stores API keys here)
+  if (!apiKeyPresent && agent.encryptedEnvVars) {
+    const authEnvVar = provider.authEnvVar;
+    if (authEnvVar && agent.encryptedEnvVars[authEnvVar]) {
```
kumaakh: The stored-encrypted-env-var fallback checks `agent.encryptedEnvVars[provider.authEnvVar]`, a static `'ANTHROPIC_API_KEY'`, but `provisionApiKey` (`src/tools/provision-auth.ts:162`, via `provider.authEnvVarForToken`) stores non-`sk-ant-` tokens under `CLAUDE_CODE_OAUTH_TOKEN` instead.

**Failure scenario:** A member provisioned via a Claude Code OAuth-style token (any token not starting with `sk-ant-`) has working credentials in `encryptedEnvVars.CLAUDE_CODE_OAUTH_TOKEN` and the remote env var of the same name, both actually used by the CLI dispatch. `preflightCheck` only ever looks up `ANTHROPIC_API_KEY` (both in the remote apiKeyCheck exec and in the encryptedEnvVars fallback), finds nothing, and with no OAuth file present returns `ok:false, code:'auth_missing'` -- rejecting every `execute_prompt` dispatch to a member whose credentials actually work.

src/tools/execute-prompt.ts
```diff
@@ -599,8 +627,8 @@ export async function executePrompt(input: ExecutePromptInput, extra?: any): Pro
   // gives it basic MCP tool access, apra-fleet-fnz.1-3) without that meaning
   // it can receive or act on this push -- routing to it anyway would silently
   // spend the full timeout_s waiting for a response that can never arrive.
-  const workspaceId = getTokenIssuer().workspaceId();
-  const rawSession = sessionRegistry.get(workspaceId, agent.id);
+  const workspaceId = earlyWorkspaceId;
```
kumaakh: `earlySession`/`earlyWorkspaceId` are snapshotted at line 578-579 before the new `await preflightCheck(agent)` (line 591, up to ~10-20s on a cache miss: 10s connectivity race plus parallel 10s OAuth/API-key execs), then reused verbatim at line 630-631 for the interactive-routing decision instead of re-querying `sessionRegistry`, contradicting the adjacent comment (line 603-604) that claims to check a session "connected right now."

**Failure scenario:** A member with no live channel at call start (`isChannelCapable=false`, so the full preflight await runs) completes its channel-capable MCP handshake via `sessionRegistry.register()` during that ~10-20s window. `rawSession` still reflects the pre-handshake snapshot, so `interactiveSession` is computed as `undefined` and the dispatch falls through to the slower subprocess/SSH path instead of the now-available interactive push channel -- a staleness window that did not meaningfully exist before this PR inserted a long-running await between the snapshot and its use.

src/services/preflight-check.ts
```diff
@@ -0,0 +1,225 @@
...
+  if (oauthResult && oauthResult.code === 0) {
+    const content = oauthResult.stdout.trim();
+    // readRemoteJson returns '{}' when file is missing
+    oauthFilePresent = content.length > 2 && content !== '{}';
+
+    if (oauthFilePresent) {
+      const cs = validateCredentials(content);
+      if (cs) {
+        credentialStatus = cs;
+        if (cs.status === 'expired-no-refresh') {
```
kumaakh, line 245: When the OAuth credential file is expired with no refresh token, `preflightCheck` returns `code:'auth_expired'` immediately without checking `apiKeyResult`, which was already fetched in the same `Promise.all` batch and is processed only in code that comes after this early return.

**Failure scenario:** A member has a stale/expired OAuth credentials file plus a working `ANTHROPIC_API_KEY` (env var or encryptedEnvVars) that the real CLI dispatch could use successfully. `preflightCheck` still blocks the dispatch with "expired (no refresh token), run /login" even though `apiKeyPresent` would have been true if the code had reached that check -- a working fallback credential is ignored because of branch ordering.

src/tools/execute-prompt.ts
```diff
@@ -572,6 +573,33 @@ export async function executePrompt(input: ExecutePromptInput, extra?: any): Pro
...
+  // their own liveness probes). The check is cached for 60s so
+  // back-to-back dispatches do not add latency.
+  if (agent.agentType !== 'local' && !isChannelCapable) {
+    const preflight = await preflightCheck(agent);
+    if (!preflight.ok) {
+      return {
+        text: `[FAIL] Pre-dispatch check failed for "${agent.friendlyName}": ${preflight.reason}`,
+        structuredContent: {
+          isError: true,
+          reason: preflight.code === 'offline' ? 'dispatch_failed' : 'auth',
```
kumaakh: The single line `reason: preflight.code === 'offline' ? 'dispatch_failed' : 'auth'` collapses distinct preflight failure modes into reason codes that fleet-sprint's `errors.mjs` (`packages/apra-fleet-se/fleet-sprint/errors.mjs:363-437`) already treats very differently for dispatches that DID reach the CLI.

**Failure scenario:** (a) `offline` maps to `'dispatch_failed'`, which `isInfraDispatchFailure()` treats as an infra fault to retry via session-resume and eventually mark INCONCLUSIVE -- but a preflight rejection means no CLI dispatch and no session were ever created, so "resuming" is meaningless. (b) `auth_missing` and `auth_expired`-no-refresh both map to `'auth'`, which `isAuthDispatchError()` triggers self-heal via `provision_llm_auth` for -- but per this PR's own preflight-check.ts comments, an expired-no-refresh token cannot be fixed by re-provisioning (it needs an interactive `/login` first), so the self-heal path re-copies the same broken credential and the sprint loops retrying an unfixable failure instead of surfacing it to an operator.

src/tools/update-member.ts
```diff
@@ -229,6 +230,15 @@ export async function updateMember(input: UpdateMemberInput): Promise<string> {
   logLine('update_member', `id=${updated.id} name=${updated.friendlyName}`, updated);
   writeStatusline();
 
+  // Invalidate preflight cache when connection or credential identity changes
+  const identityChanged = hostChanged || portChanged ||
```
kumaakh, line 234: `identityChanged` (which gates `invalidatePreflightCache`) omits `input.username`, even though `if (input.username) updates.username = input.username;` (line 195) is applied and username is part of the SSH connection identity (host/port/username) that `strategy.testConnection()` actually authenticates against.

**Failure scenario:** `update_member` changes only a remote member's username (e.g. repointing to a different login account). The 60s preflight cache is not invalidated, so any dispatch within that window returns a stale cached `ok` for the pre-change username without re-verifying the new one can actually connect or has credentials.

src/services/preflight-check.ts
```diff
@@ -0,0 +1,225 @@
...
+  let latencyMs: number;
+  let connectivityTimer: ReturnType<typeof setTimeout> | undefined;
+  try {
+    const conn = await Promise.race([
+      strategy.testConnection(),
+      new Promise<never>((_, reject) => {
+        connectivityTimer = setTimeout(() => reject(new Error('preflight connectivity timeout (10s)')), 10_000);
+      }),
+    ]);
+    clearTimeout(connectivityTimer);
```
kumaakh, line 138: `Promise.race([strategy.testConnection(), timeoutPromise])` resolves as soon as the 10s timer fires, but the underlying `testConnection()` call is never cancelled or given a `.catch`, so it keeps running after `preflightCheck` has already returned an offline result to its caller.

**Failure scenario:** Against a slow/degraded link, a `testConnection()` call that eventually resolves or rejects after the 10s race timeout has no attached handler once the race settles via the timeout branch, risking an unhandled promise rejection, and any pool/strategy-state mutation it performs on completion races with the next `preflightCheck` call for the same agent.

src/services/preflight-check.ts
```diff
@@ -0,0 +1,225 @@
...
+  /** Credential freshness detail (only for OAuth providers) */
+  credentialStatus?: CredentialStatus;
+  /** Latency of the connectivity check in ms */
+  latencyMs?: number;
+}
+
+/** Members that passed preflight within the last CACHE_TTL_MS do not need a
+ *  recheck -- avoids adding round-trip latency to back-to-back dispatches. */
+const CACHE_TTL_MS = 60_000; // 1 minute
+const preflightCache = new Map<string, { passedAt: number }>();
```
kumaakh, line 53: The module-level `preflightCache` Map has no TTL sweep or size cap; entries are only removed by `invalidatePreflightCache(id)` or a full `clearPreflightCache()`, so members whose IDs are churned through a path other than the four patched call sites (or removed/re-created repeatedly in dynamic/CI fleets) leave orphaned `${id}:conn` / `${id}:full` entries forever.

**Failure scenario:** A long-running fleet server that creates and destroys many ephemeral members over time (dynamic worker pools, CI-driven member churn) accumulates two Map entries per churned member id indefinitely -- a slow, unbounded memory leak only cleared by a process restart.

kumaakh:
**Additional finding (outside this PR's diff, so posted as a general comment rather than inline):**

`src/services/cloud/lifecycle.ts` -- Cloud member IP-drift/reattach handling (`updateAgent(agent.id, { host: currentIp })` at line 96 and `{ host: newIp }` at line 126) changes `agent.host` but never calls `invalidatePreflightCache`, unlike the four call sites this PR patched in `provision-auth.ts`/`update-member.ts`/`remove-member.ts`. The same missing-invalidation gap exists in `src/tools/setup-ssh-key.ts:116`, which rotates `authType`/`keyPath` via `updateAgent` without invalidating the cache either.

**Failure scenario:** A cloud member is preflight-checked and cached `ok` against its old IP; `ensureCloudReady()` then rotates the host to a new IP on instance restart/reattach. Within the 60s cache window, `execute_prompt`/`execute_command` skip re-verification and dispatch against the new host without confirming reachability or credentials -- exactly the class of ephemeral/reattaching member this feature was built to protect against.

omkarsm:
@kumaakh All 8 round-2 findings addressed:

1. **Cache invalidation (lifecycle/SSH key)** -- added `invalidatePreflightCache` to `cloud/lifecycle.ts` (IP rotation/reattach) and `setup-ssh-key.ts` (auth type change)
2. **CLAUDE_CODE_OAUTH_TOKEN** -- now probed via `provider.authEnvVarForToken()`; also filters empty/invalid env var names to prevent synchronous throw for OpenCode-like providers (regression caught and fixed in follow-up)
3. **Session re-query** -- `sessionRegistry` re-queried after preflight await
4. **API key fallback** -- `apiKeyResult` checked before early-returning on expired OAuth
5. **Reason codes** -- distinct `preflight_offline`/`preflight_auth_missing`/`preflight_auth_expired` with matching `errors.mjs` classification
6. **Username gate** -- added to `identityChanged` in `update-member.ts`
7. **Unhandled rejection** -- `.catch(() => {})` on `testConnection` after timeout race
8. **Cache size cap** -- CACHE_MAX_SIZE=500 with oldest-entry eviction

Added comprehensive test coverage across 7 new/updated test files including an `errors.mjs` truth table. CI green on all 3 platforms.

src/services/preflight-check.ts
```diff
@@ -0,0 +1,285 @@
...
+
+  // Process OAuth result
+  if (oauthResult && oauthResult.code === 0) {
+    const content = oauthResult.stdout.trim();
+    // readRemoteJson returns '{}' when file is missing
+    oauthFilePresent = content.length > 2 && content !== '{}';
+
+    if (oauthFilePresent) {
+      const cs = validateCredentials(content);
+      if (cs) {
```
kumaakh, line 243: **Finding: malformed/unrecognized Claude OAuth credential file silently passes preflight without any freshness check**

When `provider.name === 'claude'` and `validateCredentials(content)` returns `null` (JSON parse failure, or the expected `claudeAiOauth.expiresAt` key missing/malformed -- see `src/utils/credential-validation.ts:9-14`), neither branch at line 234 (`if (cs)`) nor the line 256 `else if (provider.name !== 'claude')` fires. `oauthFilePresent` is already `true` (set at line 230 from a non-empty, non-`'{}'` file), so the function falls straight through to `return { ok: true, ... }` without ever validating token freshness for the very provider this check exists for.

**Failure scenario**: a Claude member's `.credentials.json` gets corrupted (partial write, wrong schema, interrupted `/login`). `validateCredentials` returns `null`, preflight reports `ok: true`, and the actual dispatch then fails on real auth after the round trip this feature was built to avoid.

**Suggested fix**: treat `cs === null` the same regardless of provider -- log a warning and skip the freshness gate (don't silently pass):
```ts
if (cs) {
  ...
} else {
  logLine('preflight', `OAuth file present but unparseable for provider ${provider.name}, skipping freshness check`, agent);
}
```
so a malformed Claude credential file gets the same explicit "skipping" signal instead of silently reporting healthy.

src/tools/execute-command.ts
```diff
@@ -144,6 +147,25 @@ export async function executeCommand(input: ExecuteCommandInput, extra?: any): P
...
+  // member is reachable before attempting the command. Skips LLM auth check
+  // since execute_command runs shell commands, not LLM prompts.
+  if (agent.agentType !== 'local') {
+    const preflight = await preflightCheck(agent, { skipAuth: true });
+    if (!preflight.ok) {
+      return {
+        text: `[FAIL] Pre-dispatch check failed for "${agent.friendlyName}": ${preflight.reason}`,
+        structuredContent: {
+          isError: true,
+          reason: 'preflight_failed',
```
kumaakh: **Finding: execute_command's preflight failure uses a generic `reason: 'preflight_failed'` that none of the updated errors.mjs classifiers recognize**

`execute-prompt.ts` maps its preflight failures to specific codes (`preflight_offline`, `preflight_auth_missing`, `preflight_auth_expired`) and `packages/apra-fleet-se/fleet-sprint/errors.mjs` was updated in this same PR to route those three codes through `isNonRetryableDispatchError`, `isAuthDispatchError`, and `INFRA_DISPATCH_REASONS`. `execute-command.ts`'s own preflight failure here instead always reports the single generic `reason: 'preflight_failed'` (see line 160), which does not match any branch in any of the three classifiers -- it isn't `'auth'`, `'workspace_not_trusted'`, or any `preflight_*` code, and it isn't in `INFRA_DISPATCH_REASONS`.

**Failure scenario**: a member goes offline and a caller dispatches `execute_command` against it (e.g. fleet-sprint's git/status probes). The resulting error carries `reason: 'preflight_failed'` with a message like `Member "x" is unreachable: ...`, which will fall through to the classifiers' regex fallback (`NON_RETRYABLE_DISPATCH_RE` / `AUTH_DISPATCH_RE`) and won't match "is unreachable" either -- so a genuinely offline member's `execute_command` failures are silently mis-classified as neither infra nor auth nor non-retryable, unlike the equivalent `execute_prompt` case this PR explicitly fixed.

**Suggested fix**: reuse the same `preflight.code`-to-reason mapping used in `execute-prompt.ts` (`preflight_offline` / `preflight_auth_missing` / `preflight_auth_expired`) here instead of the single `'preflight_failed'` literal, so both dispatch paths report consistent, classifiable reason codes.

src/services/preflight-check.ts
```diff
@@ -0,0 +1,285 @@
...
+import { logLine } from '../utils/log-helpers.js';
+
+export interface PreflightResult {
+  ok: boolean;
+  /** Connectivity test passed */
+  connectivity: boolean;
+  /** LLM auth appears valid (credential files present + not expired) */
+  authValid: boolean;
+  /** Human-readable explanation when ok=false */
+  reason?: string;
```
kumaakh, line 41: **Cleanup: `CACHE_MAX_SIZE = 500` caps *entries*, not *members*, so the doc comment overstates real headroom**

The cache is keyed `${agent.id}:conn` and `${agent.id}:full` (two possible keys per member). `execute-command.ts` always calls with `{ skipAuth: true }` (populates `:conn`) while `execute-prompt.ts` calls with the default (populates `:full`). A member dispatched through both tools occupies two entries, so the comment above `CACHE_MAX_SIZE` ("A fleet with 500 members ... would otherwise grow the Map unboundedly") overstates real capacity by up to 2x -- with mixed usage, `evictOldestEntries()`'s O(n log n) full-map sort starts firing around ~250 active members, not 500.

**Suggested fix**: either track member count directly (e.g. a parallel `Set<memberId>` or count by `agent.id` prefix) or update the comment to state the cap is per cache-entry (2 possible entries/member) rather than per-member, so the eviction threshold is accurately documented.

src/services/preflight-check.ts
```diff
@@ -0,0 +1,285 @@
...
+/** Members that passed preflight within the last CACHE_TTL_MS do not need a
+ *  recheck -- avoids adding round-trip latency to back-to-back dispatches. */
+const CACHE_TTL_MS = 60_000; // 1 minute
+const preflightCache = new Map<string, { passedAt: number }>();
+
+/** R2-F8: hard cap on cache size. When exceeded, the oldest entries (by
+ *  passedAt) are evicted down to 80% of the cap. A fleet with 500 members
+ *  running back-to-back dispatches would otherwise grow the Map unboundedly
+ *  (entries only expire passively on lookup, never proactively). */
+const CACHE_MAX_SIZE = 500;
```
kumaakh, line 62: Altitude: cache invalidation is hand-added at 5 separate call sites instead of at the single choke point they all already share.

This PR wires invalidatePreflightCache(agent.id) into provision-auth.ts, remove-member.ts, update-member.ts, setup-ssh-key.ts, and cloud/lifecycle.ts individually. provision-auth.ts:178, setup-ssh-key.ts:117, and cloud/lifecycle.ts:97/128 already call updateAgent(agent.id, {...}) from src/services/registry.ts to persist the same state change that now also needs a cache invalidation; remove-member.ts similarly calls removeFromRegistry(agent.id). update-member.ts is the one caller that reasonably needs field-level filtering (identityChanged), since not every update should evict the cache.

Failure scenario: this pattern is fragile going forward -- a future call site that mutates agent host/credentials/auth-type and calls updateAgent() (nothing enforces the pairing) will silently forget the accompanying invalidatePreflightCache() call, so preflight keeps reporting the member ready off a stale cached pass for up to 60s after a credential rotation, SSH key change, or IP change.

Suggested fix: for the 4 non-update-member call sites (which always want unconditional invalidation on the identity field they touch), consider having updateAgent() itself detect host/credential-bearing field changes and call invalidatePreflightCache internally, so new callers get the invalidation for free instead of having to remember it. update-member.ts's conditional identityChanged gate can stay a special case since it genuinely needs to be selective.

src/services/preflight-check.ts
kumaakh: Fixed in 35f4b458: cs === null now always logs + skips the freshness gate explicitly (unconditional, not gated on provider name), matching the suggested fix exactly. A malformed/unparseable Claude credential file no longer falls through to ok:true.

src/tools/execute-command.ts
kumaakh: Fixed in 35f4b458: execute-command.ts now reuses the same preflight.code-to-reason mapping execute-prompt.ts uses (preflight_offline/preflight_auth_missing/preflight_auth_expired) instead of the generic preflight_failed literal, so errors.mjs classifiers recognize execute_command preflight failures the same way they recognize execute_prompt ones.

src/services/preflight-check.ts
kumaakh: Fixed in 35f4b458: corrected the doc comment above CACHE_MAX_SIZE to state the cap is per cache-entry (up to 2 keys per member), not per-member. Went with the doc-fix option per your suggested fix.

src/services/preflight-check.ts
kumaakh: Not addressed in this batch (35f4b458/b7b5f2c6) -- this is a real design suggestion but a structural refactor (folding invalidatePreflightCache into updateAgent/removeFromRegistry) is out of scope alongside the other fixes here, and touches call sites beyond what those commits changed. Leaving open as a legitimate follow-up rather than closing it out.

kumaakh:
Independent regression review (fresh pass, not a re-litigation of the two prior rounds) found 5 additional issues plus 3 of the 4 still-open threads from 2026-08-14 -- fixed in `35f4b458`/`b7b5f2c6`:

1. **Probe-failure vs genuine missing-creds**: a preflight probe that errors out (SSH hiccup, transient exec failure) no longer reports `auth_missing` -- that was misclassified as non-retryable and triggered auth self-heal on a member whose credentials may be fine. Fails open when every attempted probe errored.
2. **Connectivity timeout** raised 10s -> 30s to cut false-positive `preflight_offline` on slow-but-healthy members.
3. **Malformed Claude OAuth credential file** no longer silently passes (was falling through to `ok:true`).
4. **`execute-command.ts`** now maps preflight failures through the same reason-code logic as `execute-prompt.ts` instead of the unclassified `preflight_failed` literal.
5. **Busy-lock race**: `inFlightAgents.add()` moved to before the preflight await (was ~150 lines later), closing a window where two near-simultaneous dispatches to the same member could both pass the busy check and double-dispatch. A follow-up independent review of this specific change found one more real gap (lock leak if `preflightCheck` itself throws instead of resolving `{ok:false}`) -- fixed in `b7b5f2c6` with a regression test.
6. Kill-switch: `APRA_FLEET_PREFLIGHT=0` disables the gate fleet-wide without a code revert.
7. Verified `packages/apra-fleet-client` needs no changes -- it's a pure passthrough with no reason-code mirror for any dispatch failure code, old or new, so there was nothing to sync despite CLAUDE.md's general rule appearing to apply at first glance.

3 of the 4 2026-08-14 threads resolved with fix references inline. The 4th (cache-invalidation choke-point refactor) is left open as a legitimate design follow-up, not addressed in this batch.

143/143 tests pass (build + preflight-check/execute-prompt/execute-command suites). CI running fresh on `b7b5f2c6`.

src/services/preflight-check.ts
kumaakh: Fixed: preflightCheck now reads provider.authEnvVarForToken to resolve the correct encrypted env var (e.g. CLAUDE_CODE_OAUTH_TOKEN) instead of a hardcoded name. Confirmed present in src/services/preflight-check.ts on the current head commit.

src/tools/execute-prompt.ts
kumaakh: Fixed: preflightCheck now re-queries the session instead of relying on an early snapshot, so a session established after the initial read is still picked up. Confirmed present in src/services/preflight-check.ts on the current head commit.

src/services/preflight-check.ts
kumaakh: Fixed: preflight no longer reports auth_expired without checking whether a working api-key fallback exists first -- if the api-key probe resolves cleanly, that counts as a pass. Confirmed present in src/services/preflight-check.ts (commit range 900d7cb9..332ede5a, and further hardened in 93045703 for the per-probe indeterminacy case).

src/tools/execute-prompt.ts
kumaakh: Fixed: execute-prompt.ts and execute-command.ts now map preflight failures to distinct reason codes (preflight_offline / preflight_auth_expired / preflight_auth_missing) instead of collapsing them into one generic reason. Confirmed present in src/tools/execute-prompt.ts and src/tools/execute-command.ts on the current head commit; covered by tests/execute-command.test.ts.

src/tools/update-member.ts
kumaakh: Fixed: update-member.ts now includes input.username !== undefined in the identityChanged gate, so a username change correctly invalidates the preflight cache instead of leaving a stale entry keyed to the old identity. Confirmed present in src/tools/update-member.ts (identityChanged, lines ~234-235) on the current head commit.

src/services/preflight-check.ts
kumaakh: Fixed: testConnection now clears its timeout on both the success and catch paths, so it no longer leaves a dangling timer (and unhandled-rejection risk) behind after a completed probe. Confirmed present in the relevant strategy/connection code (commit 332ede5a) on the current head commit.

src/services/preflight-check.ts
kumaakh: Fixed: the preflight cache is now capped via CACHE_MAX_SIZE=500 with an evictOldestEntries step, and the doc comment was corrected to state it caps total entries (up to 2 per member: :conn and :full keys), not member count. Confirmed present in src/services/preflight-check.ts (commit 332ede5a) on the current head commit.

PR #392, opened 2026-08-10
fix(execute-prompt): survive MCP transport drops without killing remote sessions

omkarsm:
## Summary

When MCP HTTP transport drops during a long `execute_prompt` or `execute_command` call, the abort handler was killing the remote developer session via SSH. This fix makes remote sessions survive transport drops.

### Changes

**execute-prompt.ts:**
- Abort handler no longer calls `tryKillPid` -- only logs the disconnect via `scope.abort()`
- `dispatchSignal` uses only `stallAbortController.signal`, excluding the MCP client signal from the SSH abort chain
- `stop_prompt` explicit user-initiated kill path remains unchanged

**execute-command.ts:**
- Same abort-fix pattern applied: abort handler logs only, no `tryKillPid` call

**dispatch-helpers.ts (new):**
- Extracted shared `attachMcpDisconnectHandler()` to avoid duplication between execute-prompt and execute-command

**prompt_progress reverted:**
- Reverted `95992f3` per reviewer feedback -- fleet-mcp layer must not assume git/progress.json workflows (layering violation)
- Progress monitoring will be reimplemented in the pm skill layer

### Validation

Tested in production: 3 consecutive MCP transport drops survived during a 90-minute EVIDENT D-64 dispatch. Previously every transport drop killed the remote session.

## Review feedback addressed (round 1)

1. `execute-command.ts` had the same premature-kill bug -- fixed with identical abort pattern

## Review feedback addressed (round 2)

3 additional findings from @kumaakh's second review:

1. **R2-A1**: Removed stale `extra?.signal?.aborted` retry-skip guard (no longer meaningful after dispatchSignal change)
2. **R2-A2**: Simplified stall disambiguation to just `stallAbortController.signal.aborted` (extra.signal exclusion makes second condition redundant)
3. **R2-A3**: Extracted shared MCP disconnect handler to `src/services/dispatch-helpers.ts` (was duplicated verbatim in execute-command and execute-prompt)

## Test plan

- [x] T10 test suite updated for new MCP disconnect behavior
- [x] Dispatch-exception retry test updated for d64.1 signal change
- [x] 111/111 tests pass across execute-prompt and execute-command test files
- [x] TypeScript build passes
- [x] CI green


kumaakh:
@omkarsm prompt progress tracking is a good requirement for sure. 
But we can not create a fleet tool which assumes that the progress is measure by git changes or progress.json

This is a right feature to add to  /pm (pm skill) and fleet-sprint workflow but not in fleet-mcp.

As this will be a violation of layering principal... such lower layer can not assume how it is being used by a higher layer...
there may be a 1000 other ways of using execture_prompt which have nothing to do with git or progress.json

since the "progress.json" concept is added by pm skill, we should create a new /pm command which is able to look at the fleet logs and member side conversation logs (like stall detector module) to report the progress.

omkarsm:
Agreed -- prompt_progress reverted in 2b66f2d. The git/progress.json monitoring belongs in the pm skill layer, not fleet-mcp. Will implement progress tracking there instead. Thanks for the feedback.

src/tools/execute-prompt.ts
```diff
@@ -1013,20 +1014,22 @@ export async function executePrompt(input: ExecutePromptInput, extra?: any): Pro
     }
   };
 
+  // apra-fleet-d64.1: MCP transport drops must NOT kill the remote process.
+  // The remote CLI session is independent of the MCP session and should
+  // continue working. The abort handler only logs the disconnection --
+  // tryKillPid is NOT called here. Explicit kills go through stop_prompt;
+  // stall kills go through the stallDetector's onStall callback (line ~715).
   const abortHandler = () => {
```
kumaakh: **[correctness]** Dropping `extra.signal` from `dispatchSignal` conflates two distinct MCP signals: a raw transport disconnect and an explicit per-request `CancelledNotification` from the client both fire the same request-handler `AbortController` (per the MCP SDK: a `CancelledNotification` handler calls `controller.abort(reason)`, and `_onclose()` also aborts all pending controllers on transport close -- same signal, two different causes).

**Failure scenario:** a client that is still connected sends a legitimate MCP cancel for this specific `execute_prompt` request (e.g. user hits stop in a UI that cancels the pending tool call, not a network drop). Before this change, that killed the remote CLI process immediately. Now `abortHandler` only logs and `dispatchSignal` no longer observes `extra.signal`, so the remote process keeps running until natural completion or a stall is detected -- an explicit, still-connected cancellation no longer stops the remote work. The only documented remaining kill path is the separate `stop_prompt` tool, which the disconnected caller must now know to invoke.

src/tools/execute-prompt.ts
```diff
@@ -298,7 +298,8 @@ async function ensureAgentFilesProvisioned(agent: Agent): Promise<void> {
 // (a) normal success: result.code === 0 -> finally sets idle and removes agent from inFlight
 // (b) non-zero exit from execCommand: result.code !== 0 -> finally sets idle and removes agent from inFlight
 // (c) exception in try block (auth, network, crash) -> catch records error type; finally sets offline or idle
-// (d) AbortSignal/MCP client cancellation -> abortHandler kills PID, execCommand resolves, finally clears
+// (d) MCP client disconnection -> remote process continues; SSH stays alive until natural completion or timer
+//     Stall abort -> abortHandler kills PID via onStall, execCommand resolves, finally clears
```
kumaakh, line 303: **[simplification/doc-accuracy]** This scenario comment (d) misattributes the stall-triggered PID kill to `abortHandler`, but `abortHandler` (line ~1022) now only logs -- the kill is actually done by the unrelated `onStall` callback registered on the stall detector (around line 715-719), not by `abortHandler`.

**Failure scenario:** a future maintainer reads this (a)-(g) summary to understand cleanup paths, sees "Stall abort -> abortHandler kills PID via onStall", concludes `abortHandler` still performs kills on some path, and wires new abort-based cleanup through `abortHandler` instead of `onStall` -- reintroducing the kill-on-disconnect bug this PR just removed.

kumaakh:
**[correctness]** `src/tools/execute-command.ts:153` (not touched by this PR's diff, so I can't attach this inline): this file still has the pre-fix `abortHandler` that calls `tryKillPid` on `extra.signal` abort -- the identical premature-kill-on-MCP-disconnect problem that apra-fleet-d64.1 fixed here for `execute_prompt` (see `execute-prompt.ts:1022`, which now only logs).

**Failure scenario:** an MCP transport drop during a long-running `execute_command` dispatch still kills the remote process via `tryKillPid(agent, strategy, cmds)`, exactly the behavior this PR concluded was wrong for `execute_prompt`. The fix was special-cased to one tool instead of the shared abort-wiring/strategy layer, leaving `execute_command`'s remote commands non-resilient to the same transient disconnects `execute_prompt` is now resilient to. Worth a follow-up bead to apply the same fix there.

omkarsm:
@kumaakh Good catch on `execute-command.ts` -- applied the same abort-fix pattern in `f2de2165`:

- Removed `tryKillPid` from abort handler (now logs only, same as execute-prompt.ts)
- Removed unused `tryKillPid` import
- Added test verifying no remote kill on MCP disconnect
- `stop_prompt` explicit kill path unchanged

Both `execute_prompt` and `execute_command` are now resilient to MCP transport drops. CI passing on ubuntu/macos, windows pending.

kumaakh (commented):
Automated re-review (fresh pass against the current/reworked diff).

src/tools/execute-command.ts
```diff
@@ -150,9 +149,12 @@ export async function executeCommand(input: ExecuteCommandInput, extra?: any): P
 
   const cmds = getOsCommands(getAgentOS(agent));
   const agentOs = getAgentOS(agent);
+    // apra-fleet-d64.1: MCP transport drops must NOT kill the remote process.
+    // The remote command is independent of the MCP session and should continue
+    // running. The abort handler only logs the disconnection -- tryKillPid is
+    // NOT called here. Same pattern as execute-prompt.ts.
     const abortHandler = () => {
```
kumaakh: The `abortHandler` closure (log-only disconnect message, addEventListener/removeEventListener pair) is now duplicated verbatim between `execute-command.ts` and `execute-prompt.ts`, with the new comment in `execute-command.ts` explicitly noting "Same pattern as execute-prompt.ts."

**Impact:** Any future change to the disconnect message or behavior (e.g. adding a metric, changing the log level) must be kept in sync by hand across both files; a small shared helper (e.g. in `log-helpers.ts`) would remove the copy-paste the diff itself calls out.

kumaakh:
**2 additional findings, both outside this PR's diff hunk (so posted as a general comment rather than inline) -- pre-existing code that's now broken by the \`dispatchSignal\` semantic change at lines 1013-1035:**

**1. \`src/tools/execute-prompt.ts:1060\`** -- Retry-skip guard \`if (extra?.signal?.aborted) throw dispatchErr;\` still assumes an MCP disconnect caused the dispatch exception, but \`dispatchSignal\` no longer includes \`extra.signal\` (it's now just \`stallAbortController.signal\`, per this PR's own design).

**Failure scenario:** Client disconnects mid-dispatch (\`extra.signal\` fires abort, but per this PR's design that no longer touches \`dispatchSignal\`/the SSH channel). Seconds later \`strategy.execCommand\` throws for an unrelated reason (e.g. \`'Command timed out after Xms of inactivity'\` from \`src/services/ssh.ts\`, or any SSH/stream error) -- these throws are independent of \`extra.signal\`. Because \`extra.signal.aborted\` is now \`true\` (just from the earlier disconnect), the fresh-session retry recovery is skipped and the exception is rethrown as terminal, even though nothing about the failure was actually caused by the client leaving. This directly undermines the PR's stated goal that a transport drop must not change dispatch/recovery behavior.

**2. \`src/tools/execute-prompt.ts:1385\`** -- Outer catch guard \`if (stallAbortController.signal.aborted && !extra?.signal?.aborted)\` still disambiguates "stall vs client signal" as if both could jointly abort \`dispatchSignal\` -- but \`dispatchSignal\` is now only \`stallAbortController.signal\`, so \`extra.signal\` can never be the cause.

**Failure scenario:** A confirmed stall fires (\`stallAbortController.signal.aborted = true\`) around the same time the MCP client happens to have disconnected earlier in the call (\`extra.signal.aborted = true\`, an unrelated coincidence). The guard evaluates false, so the genuine stall is misclassified as generic \`dispatch_failed\` instead of the typed \`reason:'stalled'\`, losing the stall-specific message and any caller logic (e.g. stall-aware retry/backoff) keyed on that reason.

omkarsm:
@kumaakh All 3 round-2 findings addressed:

1. **Stale signal guard** -- removed `extra?.signal?.aborted` retry-skip guard (no longer meaningful after dispatchSignal change)
2. **Stall disambiguation** -- simplified to just `stallAbortController.signal.aborted`
3. **Shared handler** -- extracted `attachMcpDisconnectHandler()` to `src/services/dispatch-helpers.ts`, used by both execute-prompt and execute-command with cleanup in `finally`

Tests updated: dispatch-exception retry test reflects new signal behavior. 111/111 passing. CI green.

src/services/dispatch-helpers.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Returns a cleanup function that removes the event listener. Call it in
+ * a finally block to avoid leaking listeners on the signal.
+ */
+export function attachMcpDisconnectHandler(
+  signal: AbortSignal | undefined,
+  scope: LogScope,
+): () => void {
+  if (!signal) return () => {};
+  const handler = () => {
+    scope.abort('MCP client disconnected -- remote session continues');
```
kumaakh, line 20: **Severity-mislabeled disconnect log — contradicts this function's own docstring**

`attachMcpDisconnectHandler`'s docstring says: "The handler only logs the disconnection -- the remote session continues working independently." But the handler calls `scope.abort(...)`, which is `LogScope`'s ERROR-level, *scope-exiting* method (`_exit('error', msg)` -> `[fleet:error]` in the JSON log stream written to `fleet-<pid>.log`).

Concretely: every MCP transport drop mid-dispatch (the exact case this PR exists to make benign) now emits a `{level:'error', tag:'execute_prompt', msg:'MCP client disconnected -- remote session continues elapsed=...ms'}` line for what is explicitly *not* an error, and the same `LogScope` later emits a second, real terminal line via `scope.ok()`/`scope.fail()`/`scope.abort()` in the `finally` block once the dispatch actually settles — i.e. two "exit"-shaped log lines for one logical dispatch, the first mislabeled. Any log-scraping alert/dashboard keyed on `level: error` would false-positive on a normal, successful dispatch that merely survived a transient transport drop.

`LogScope` already has a non-exit informational method for exactly this: `scope.info(...)`.

```suggestion
    scope.info('MCP client disconnected -- remote session continues');
```

PR #398, opened 2026-08-11
feat(email): credential store + inline config for send_email

dsiddharth2:
## What this does

Adds a `send_email` MCP tool to apra-fleet that sends email via SendGrid or SMTP. The design follows a zero-env-var principle: secrets come from the fleet credential store, non-secret config is passed inline by the caller.

## How it works

**Secrets (API keys, passwords)** are stored once via the credential store:

```bash
apra-fleet secret --set smtp_password --persist
```

**Non-secret config (host, port, user, from)** is passed by the workflow in every `send_email` call. The workflow loads this config however it wants -- JSON file, hardcoded, etc.

**The workflow never sees secrets.** The server resolves them from the credential store at send time.

### Core implementation

- `src/tools/send-email.ts` -- `send_email` tool with Zod schema, address validation, and `buildProvider()` that constructs the email provider from inline config + credential store secrets
- `src/providers/email/` -- SendGrid adapter (direct HTTP v3 API, no dependencies), SMTP adapter (raw socket, STARTTLS, AUTH LOGIN, MIME multipart -- no nodemailer dependency), and `resolveSecret()` for credential store lookups
- `src/services/tool-registry.ts` -- tool registration with description guiding LLM agents toward `credential_store_set`
- `packages/apra-fleet-client/src/client/api.mjs` -- `sendEmail()` client wrapper

### Example workflow

- `examples/workflows/email-notify/` -- complete runnable workflow that checks if the credential exists, prompts via OOB if missing, sends email. Includes CI pipeline alternative via `execute_command` + stdin injection. Config loaded from a gitignored `config.json` (no secrets).

### Documentation

- `docs/email-workflow-guide.md` -- end-to-end walkthrough: one-time setup, sending from a workflow, handling missing credentials, CI pipeline pattern, parameter reference
- `README.md` -- Email Configuration section with credential store setup, workflow example, and tool reference table

### Tests

- `tests/email-resolver.test.ts` -- 8 unit tests: provider resolution via credential store, missing credential error messages, missing SMTP fields, address validation
- `tests/email-adapters.test.ts` -- 7 unit tests: SendGrid and SMTP adapter behavior with mocked network I/O
- `tests/integration/send-email.integ.test.ts` -- real-send integration test gated on `email-test-config.json` (gitignored, no env vars)

## Verified manually

- Stored `smtp_password` via CLI, sent real email via MCP call -- received at inbox
- Deleted credential, called `send_email` -- got clear error: `"Store it with credential_store_set (name: \"smtp_password\")"`
- Ran `email-notify` workflow end to end -- connected to fleet, checked credential store, sent email

## Design principles

1. **Secrets go through `credential_store_set`** -- no exceptions, no alternatives in examples
2. **Non-secret config is passed by the caller** -- the workflow owns it, not the server environment
3. **No secrets in any example** -- not in JSON, not in .env, not in bash exports
4. **No env vars anywhere in the email path** -- the `send_email` tool is fully self-contained


src/providers/email/index.ts
```diff
@@ -0,0 +1,26 @@
+import { credentialResolve } from '../../services/credential-store.js';
+import { SendGridProvider } from './sendgrid.js';
+import { SmtpProvider } from './smtp.js';
+import type { EmailConfig, EmailProvider } from './provider.js';
+
+export type { EmailAttachment, EmailMessage, EmailSendResult, EmailProvider, EmailConfig } from './provider.js';
+
+export function resolveSecret(credentialName: string): string | undefined {
```
kumaakh: Credential-scope bypass (CONFIRMED): `resolveSecret()` hardcodes `credentialResolve(credentialName, '*')`, and `sendEmail()`/`wrapTool('send_email', ...)` in `tool-registry.ts` never thread the real calling member through. Per `credential-store.ts`, `'*'` is the fleet-operator scoping bypass — any member calling `send_email` gets `sendgrid_api_key`/`smtp_password` resolved even if that credential's `allowedMembers` restricts it to specific members. Contrast with `execute-command.ts`/`provision-vcs-auth.ts`, which pass the genuine calling-member id. Recommend threading the real member (from `extra`) into `resolveSecret`/`buildProvider` so `allowedMembers` scoping is actually enforced for email credentials.

src/providers/email/smtp.ts
```diff
@@ -0,0 +1,205 @@
...
+    if (auth?.user) {
+      const authStart = await conn.command('AUTH LOGIN');
+      if (authStart.code !== 334) throw new Error(`SMTP AUTH LOGIN failed: ${authStart.lines.join(' ')}`);
+      const userResp = await conn.command(Buffer.from(auth.user, 'utf8').toString('base64'));
+      if (userResp.code !== 334) throw new Error(`SMTP AUTH username rejected: ${userResp.lines.join(' ')}`);
+      const passResp = await conn.command(Buffer.from(auth.pass, 'utf8').toString('base64'));
+      if (passResp.code !== 235) throw new Error(`SMTP AUTH failed: ${passResp.lines.join(' ')}`);
+    }
+
+    const fromResp = await conn.command(`MAIL FROM:<${msg.from ?? from}>`);
```
kumaakh: SMTP header/command injection (CONFIRMED): `sendEmailSchema` only regex-validates `to`/`cc`/`bcc` via `EMAIL_RE`; `from` and `subject` are unchecked `z.string()` values spliced raw into the protocol stream here (`MAIL FROM:<${msg.from ?? from}>`) and at line 99 (`Subject: ${msg.subject}`). A `from`/`subject` containing `\r\n` breaks out of the SMTP command / MIME header and can inject extra recipients or headers (classic SMTP header/command injection). Neither field goes through `dotStuff` or any CRLF-stripping. Recommend rejecting/stripping CR/LF in `from` and `subject` (and any other header-bound fields) before they reach the socket.

src/services/email/index.ts
```diff
@@ -0,0 +1,79 @@
+import { credentialResolve } from '../credential-store.js';
```
kumaakh: Dead duplicate implementation (CONFIRMED, corroborated by 5 of 8 review angles): `src/services/email/{index,sendgrid,smtp,types}.ts` is a near-byte-identical parallel implementation of `src/providers/email/*`. Nothing imports `services/email` — `src/tools/send-email.ts` and `tool-registry.ts` wire up `providers/email` exclusively. Worse, this dead copy's `resolveSecret(name, envVar)` falls back to `process.env` (`SENDGRID_API_KEY`, `SMTP_PASS`, etc.), directly contradicting this PR's own "no secrets in env vars" design principle — if a future change ever wires something to this directory by mistake (plausible given the near-identical names), it silently reintroduces env-var secret handling with zero test coverage. Note the diff's own bundled beads issue `apra-fleet-6nj.1.2` is marked "closed" claiming this directory was already deleted. Recommend deleting `src/services/email/` entirely before merge.

packages/apra-fleet-client/src/client/api.mjs
```diff
@@ -163,6 +163,17 @@
  * @property {string} [member_name] - Friendly name of the member
  */
 
+/**
+ * @typedef {Object} SendEmailOptions
```
kumaakh, line 187: Stale client wrapper (CONFIRMED): per this repo's CLAUDE.md, `apra-fleet-client` must be updated in the same change as any fleet MCP tool schema change. The `SendEmailOptions` JSDoc typedef here documents only `to`/`subject`/`body`/`html`/`cc`/`bcc`/`attachments` — it omits `provider`, `from` (required — no `.optional()`/`.default()` on the server schema), `host`, `port`, `user`, `secure` entirely. Callers relying on the client's type surface get no signal that `from` is mandatory or that SMTP needs `host`/`user`. Recommend adding the missing `@property` entries so the typedef matches `sendEmailSchema` in `src/tools/send-email.ts`.

src/tools/send-email.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (!EMAIL_RE.test(addr)) errors.push(`Invalid 'cc' address: ${addr}`);
+  }
+  for (const addr of input.bcc ?? []) {
+    if (!EMAIL_RE.test(addr)) errors.push(`Invalid 'bcc' address: ${addr}`);
+  }
+
+  return errors;
+}
+
+function buildProvider(input: SendEmailInput): EmailProvider {
```
kumaakh: Triplicated validation, wrong altitude (CONFIRMED): `host`/`user` are `z.string().optional()` in the schema regardless of `provider`, so the real "smtp requires host+user" rule only lives in these runtime throws (lines 63/66) — and is re-implemented a third time in `examples/workflows/email-notify/main.mjs:35-36` (`if (config.provider === 'smtp') requiredFields.push('host', 'user')`). The MCP tool's declared schema (visible to callers) doesn't reflect the real requirement, and the rule now has to be kept in sync across 3 places. Recommend a `z.discriminatedUnion('provider', [...])` so `host`/`user` are conditionally required in the schema itself — single source of truth, self-documenting to MCP clients.

src/tools/send-email.ts
```diff
@@ -0,0 +1,111 @@
...
+  if (input.provider === 'smtp') {
+    const pass = resolveSecret('smtp_password');
```
kumaakh: Validation-ordering / hot-path efficiency (CONFIRMED): `resolveSecret('smtp_password')` (a synchronous credential-store round trip: `fs.readFileSync` + `JSON.parse` + AES-GCM decrypt, uncached) runs *before* the free `input.host`/`input.user` presence checks below it. A request that's simply missing `host`/`user` still pays for full credential resolution before failing. Recommend validating cheap required fields first, resolving the secret last — and consider caching the decrypted credential in memory (invalidated on credential_store_set/update/delete) since this now sits on the send hot path.

examples/workflows/email-notify/main.mjs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// --- Phase 4: Send email ---
+//
+// Alternative: set credentials programmatically in a CI pipeline by pulling
+// from a vault and injecting via execute_command:
+//
+//   await mcpClient.callTool('execute_command', {
+//     member_name: 'local',
+//     command: 'echo "$SECRET_FROM_VAULT" | apra-fleet secret --set smtp_password --persist -y',
```
kumaakh: Anti-pattern in example (CONFIRMED, inert — 100% commented out): this "CI pipeline alternative" models piping a secret through a shell command (`echo "$SECRET_FROM_VAULT" | apra-fleet secret --set ...`) via `execute_command`. It never executes, but as guidance it teaches exposing a secret in a command string (visible in process listings / `execute_command`'s audit-log trail) right next to this same file's stated principle that secrets never leave the credential store. Recommend removing this block or replacing it with a pattern that doesn't put the secret value in a shell command string.

src/providers/email/smtp.ts
```diff
@@ -0,0 +1,205 @@
...
+  write(data: string): void {
+    this.socket.write(data);
+  }
+
+  end(): void {
+    this.socket.end();
+  }
+}
+
+function connectSocket(host: string, port: number, secure: boolean): Promise<net.Socket | tls.TLSSocket> {
```
kumaakh: **No connect/response timeout (CONFIRMED):** `connectSocket()` (lines 75-82) and `SmtpConnection.waitForResponse()` (lines 50-52) never set a socket or response timeout. If the SMTP host is unreachable, firewalled, or slow to respond (e.g. after `STARTTLS` or `AUTH LOGIN`), `send()` — and the MCP tool call wrapping it — hangs indefinitely, leaking the socket/listener for the life of the process with no way to recover short of killing the server. Suggest `socket.setTimeout()` plus a timeout race on each `command()`/`waitForResponse()`.

kumaakh:
@dsiddharth2 pls see inline PR review  comments above

src/providers/email/index.ts
dsiddharth2: Fixed in c4b96877. resolveSecret now requires a callingMember, and send_email derives it from the MCP session: a connected member session resolves under its own member friendly name (so allowedMembers is enforced), while a caller with no registered member session (the orchestrator / fleet operator) keeps the '*' scope, same as the CLI path. Denied/expired credentials now surface their real reason instead of falling through to 'not found'. New tests cover member-session scoping, denial surfacing, and the operator fallback.

src/providers/email/smtp.ts
dsiddharth2: Fixed in c4b96877, two layers: the tool now validates 'from' against the address regex and rejects CR/LF in 'subject' before any credential or network work, and the SMTP provider independently rejects CR/LF in every header/command-bound value (from, recipients, subject, attachment filename/contentType) before any I/O. Tests cover both layers.

src/services/email/index.ts
dsiddharth2: Deleted in c4b96877 -- src/services/email/ is gone, including its env-var resolveSecret fallback. Also removed the unused getEmailProvider() from src/providers/email/index.ts in the same spirit (nothing called it).

packages/apra-fleet-client/src/client/api.mjs
dsiddharth2: Fixed in c4b96877. SendEmailOptions now documents provider, from (required), host, port, user, and secure, matching sendEmailSchema.

src/tools/send-email.ts
dsiddharth2: Partially applied in c4b96877 -- z.discriminatedUnion cannot be used here: McpServer.tool() only accepts a flat ZodRawShape (we pass sendEmailSchema.shape) and the SDK rebuilds z.object(shape) internally, so a top-level union (or even a .superRefine on the object) never survives registration. What I did instead: the host/user rule now lives in exactly one runtime place -- validateInput, next to the address checks, with a comment explaining the SDK constraint -- and the field descriptions still document the conditional requirement to MCP clients. The check in the example workflow stays as a client-side fail-fast on config.json before connecting; it validates a local config file, not the tool input, so it is a different layer rather than a third copy of the same rule.

src/tools/send-email.ts
dsiddharth2: Fixed in c4b96877 -- all cheap field checks (addresses, from, subject, smtp host/user) now run before any credential-store round trip, and tests assert credentialResolve is never called when validation fails. I deliberately skipped the in-memory credential cache: sends are low-frequency, the store read is a small local file plus one AES-GCM decrypt, and a stale-cache bug on secrets (rotation not picked up) would cost more than the read.

examples/workflows/email-notify/main.mjs
dsiddharth2: Removed in c4b96877. The guide now shows seeding the credential in a direct CI shell step, piping from the vault CLI over stdin (apra-fleet secret --set smtp_password --persist -y), with an explicit warning not to route secrets through execute_command since the command string is visible to the LLM session and recorded in the audit trail. This matches what .github/workflows/fleet-e2e.yml already does.

src/providers/email/smtp.ts
dsiddharth2: Fixed in c4b96877. connectSocket, the STARTTLS handshake, and every response wait now have a timeout (30s default, configurable via SmtpConfig.timeoutMs) that destroys the socket on expiry, so an unreachable or silent SMTP host fails the tool call instead of hanging the server. Tests cover both the connect-timeout and response-timeout paths.

dsiddharth2:
Addressed all 8 review comments in c4b96877 (replies inline on each thread).

Also verified end to end after the changes: restarted the fleet server on the new build and ran the email-notify workflow against real Gmail SMTP -- credential resolved from the store, email delivered (messageId smtp-1786548656484).

The test run surfaced one additional bug, fixed in b9eee095: the example workflow parsed credential_store_list as {credentials: [...]} when the tool returns a JSON array, so it always thought the credential was missing and re-triggered credential_store_set on every run. The example and the guide snippet now check the array shape, and the example locates the JSON payload in tool results instead of assuming content[0] (which can be an onboarding display block).

examples/workflows/hello-world/main.mjs
```diff
@@ -26,4 +26,28 @@ const engine = WorkflowEngine ? 'resolved' : 'missing';
...
+  .catch(() => ({}));
+
+if (fleetApi) {
+  const emailResult = await fleetApi.sendEmail({
+    provider: 'sendgrid',
+    from: 'hello@example.com',
+    to: 'team@example.com',
+    subject: 'Hello from apra-fleet workflow',
+    body: 'Sent without any secrets in this file.',
+  }).then(r => JSON.parse(r)).catch(e => ({ ok: false, error: e.message }));
```
kumaakh: **Bug**: `fleetApi.sendEmail()` (the `ApraFleet` client wrapper, `packages/apra-fleet-client/src/client/api.mjs:369`) returns the raw MCP `callTool()` result -- `{ content: [...], structuredContent? }` -- not a JSON string. `JSON.parse(r)` here will throw `SyntaxError: Unexpected token o in JSON` (parsing `"[object Object]"`), which the `.catch(e => ({ ok: false, error: e.message }))` silently swallows.

Net effect: this demo always prints `[SKIP] email: Unexpected token o in JSON...` even when SendGrid is fully configured and the send actually succeeded server-side -- it never demonstrates a working send.

Compare with `examples/workflows/email-notify/main.mjs`, which does this correctly via its `parseToolJson()` helper that scans `result.content` for the first JSON-parseable text item (tool results can have a display preamble ahead of the payload). The `.then(r => JSON.parse(r))` pattern here needs the same unwrapping.

docs/email-workflow-guide.md
```diff
@@ -0,0 +1,190 @@
...
+A workflow calls `send_email` through the fleet MCP client. Non-secret config
+is passed inline. The credential store provides the password/API key
+automatically.
+
+```javascript
+import { connectFleet } from '@apralabs/apra-fleet-client/server-resolution';
+
+const { fleetApi } = await connectFleet({ env: process.env });
+
+const result = JSON.parse(await fleetApi.sendEmail({
```
kumaakh: **Bug (docs)**: same issue as `examples/workflows/hello-world/main.mjs` -- `fleetApi.sendEmail(...)` returns the raw MCP `callTool()` result (`{ content: [...] }`), not a JSON string. `JSON.parse(await fleetApi.sendEmail({...}))` here will throw at runtime; a workflow author following this guide verbatim gets a crash, not the `if (result.ok) {...}` branch shown below it.

Needs the same unwrap `examples/workflows/email-notify/main.mjs` uses (pull the JSON payload out of `result.content`, skipping any `<apra-fleet-display>` preamble item) before `JSON.parse`.

README.md
```diff
@@ -253,6 +253,87 @@ a network egress policy (allow / deny / confirm). Every member runs with
...
+    provider: 'smtp',
+    host: 'smtp.example.com',
+    port: 587,
+    user: 'notifications@example.com',
+    from: 'noreply@example.com',
+    to: 'team@example.com',
+    subject: 'Sprint Report',
+    body: 'All tasks completed.'
+  });
+  log(`Sent: ${JSON.parse(result).messageId}`);
```
kumaakh: **Bug (docs)**: same issue flagged in `examples/workflows/hello-world/main.mjs` and `docs/email-workflow-guide.md` -- `fleetApi.sendEmail()` returns the raw MCP tool-call result object (`{ content: [...] }`), not a JSON string, so `JSON.parse(result)` here throws instead of producing `{ ok, messageId }`. This is the top-level README quickstart, so its wrong; needs the same content-array unwrap that `examples/workflows/email-notify/main.mjs`s `parseToolJson()` does.

src/providers/email/smtp.ts
```diff
@@ -0,0 +1,278 @@
...
+      return;
+    }
+
+    const code = parseInt(match[1], 10);
+    const waiter = this.waiters.shift();
+    if (waiter) {
+      clearTimeout(waiter.timer);
+      waiter.resolve({ code, lines });
+    }
+  }
```
kumaakh, line 101: **Plausible (uncertain trigger)**: `onData()` assumes each invocation delivers at most one complete SMTP response, and resolves exactly one waiter per call by taking `lines[lines.length - 1]` as "the" final line. If the TCP stack ever coalesces two full server responses into a single `data` event while two `waitForResponse()` calls are outstanding (this cannot happen with the current strictly-sequential `await conn.command(...)` call pattern, but would if any future caller ever pipelines two commands before awaiting the first reply), the code reads the *last* full response line for both responses combined, resolves only the oldest waiter with the wrong `code`/`lines`, and leaves the second waiter to hang until the 30s timeout.

Would be confirmed by: adding a second call site that issues two SMTP commands without awaiting the response in between (e.g. a future pipelining optimization), or a test that feeds `onData()` two concatenated responses in one chunk while two waiters are queued.

examples/workflows/hello-world/main.mjs
dsiddharth2: Fixed in 7e20e14a, by removal: an independent pass over the branch also flagged that this email phase contradicts hello-world's own header contract ('makes no fleet-server connection' to keep the CI smoke test hermetic) -- connectFleet's stdio fallback would self-spawn a fleet server inside CI, and on a dev machine with sendgrid_api_key stored it would fire a real email to team@example.com. hello-world is restored to the hermetic version from main; email-notify remains the canonical email example. The underlying JSON.parse-on-raw-result bug is fixed everywhere else via a new parseToolJson() helper exported from @apralabs/apra-fleet-client.

docs/email-workflow-guide.md
dsiddharth2: Fixed in 7e20e14a. Added parseToolJson() to @apralabs/apra-fleet-client (scans result.content for the first JSON-parseable item, skipping display preambles) and the guide now uses it. Also added credentialStoreSet/List/Delete/Update wrappers to the client, so the guide's 'Handling missing credentials' snippet now calls methods that actually exist -- verified by running the email-notify workflow end to end (real SMTP send).

README.md
dsiddharth2: Fixed in 7e20e14a. The README example now uses parseToolJson() -- and while fixing it I found the example was also destructuring fleetApi from the workflow engine context, which the engine does not provide (it exposes agent/command/log/phase/args/budget only); the snippet now uses connectFleet() like the guide. llms-full.txt regenerated from the fixed README (it is generated by scripts/gen-llms-full.mjs and was carrying the stale broken pattern).

src/providers/email/smtp.ts
dsiddharth2: Fixed in 7e20e14a. onData() now extracts every complete response from the buffer in a loop (one waiter resolved per response, in order), and a response that completes with no waiter queued is held in a pending queue that waitForResponse() consumes first -- which also fixes the latent greeting race where a fast server response could arrive before the waiter registered. Covered by two new tests: two coalesced responses delivered to two queued waiters in order, and an early-arriving response consumed by a later wait. Your suggested trigger (pipelined commands) now works correctly if a future caller ever does it.

dsiddharth2:
Round-2 comments addressed in 7e20e14a (replies inline).

Also ran an independent review pass over the full branch diff before pushing; it surfaced and 7e20e14a fixes several additional issues beyond the inline comments:

- SMTP: message bodies are now CRLF-normalized before dot-stuffing (bare-LF bodies were rejected by strict servers like Gmail, and a body line of '.' with LF endings could slip past dotStuff and truncate DATA)
- send_email credential scoping now fails closed: an HTTP session that is not a registered member session gets a synthetic non-member identity instead of the '*' operator bypass ('*'-scoped credentials still work; member-scoped ones deny)
- SMTP: try/finally socket teardown on every send path, plus persistent error/close handlers so a mid-session ECONNRESET rejects the in-flight wait with the real error instead of hanging to the 30s timeout (or crashing on unhandled 'error')
- MIME: Date and Message-ID headers, RFC-2047 subject encoding, base64 attachment wrapping at 76 chars
- SendGrid: request timeout via AbortSignal (30s default), mirroring the SMTP ceiling
- Client: credentialStoreSet/List/Delete/Update wrappers added (parity convention)
- Docs: guide run instructions corrected (email-notify is not an installed workflow; run with node main.mjs or copy into ~/.apra-fleet/workflows)

Verified: build + email test suites pass (30 tests), and the email-notify workflow delivered a real email end to end through the updated client wrappers and SMTP path.

src/providers/email/smtp.ts
```diff
@@ -0,0 +1,375 @@
...
+    // failure, rejected recipient, timeout) -- without the finally the open
+    // TCP socket would leak on each failed send.
+    try {
+      const greeting = await conn.waitForResponse();
+      if (greeting.code !== 220) throw new Error(`SMTP server did not greet: ${greeting.lines.join(' ')}`);
+
+      let ehlo = await conn.command(`EHLO localhost`);
+      if (ehlo.code !== 250) throw new Error(`SMTP EHLO failed: ${ehlo.lines.join(' ')}`);
+
+      if (!secure && ehlo.lines.some(l => /STARTTLS/i.test(l))) {
```
kumaakh: **STARTTLS not enforced -- opportunistic-TLS downgrade lets credentials leak in cleartext**

TLS upgrade only happens when the server advertises STARTTLS: `if (!secure && ehlo.lines.some(l => /STARTTLS/i.test(l)))`. If a network attacker strips the STARTTLS line from the EHLO response (classic STARTTLS-stripping MITM), or the server just doesn't advertise it, this silently falls through to plaintext. `secure:false` is also the tool's default (sendEmailSchema), so a typical `port:587` config with no explicit `secure` is vulnerable by default.

Failure scenario: caller sends {provider:'smtp', host, port:587} (secure defaults to false) against a server/path where STARTTLS advertisement is stripped -- AUTH LOGIN then sends the base64-encoded (not encrypted) smtp_password over an unencrypted socket, leaking the stored credential to anyone on-path.

Suggest failing closed: require STARTTLS success (or secure:true) rather than silently proceeding over plaintext when the server doesn't advertise it.

src/tools/send-email.ts
```diff
@@ -0,0 +1,149 @@
...
+ * what workflow clients rely on), but member-scoped credentials are denied
+ * rather than silently bypassed.
+ */
+function resolveCallingMember(extra?: { sessionId?: string }): string {
+  const sessionId = extra?.sessionId;
+  if (!sessionId) return '*';
+  const session = sessionRegistry.findBySessionId(sessionId);
+  if (!session) return `session:${sessionId}`;
+  const agent = getAgentOrFail(session.member_id);
+  return typeof agent === 'string' ? session.member_id : agent.friendlyName;
```
kumaakh: **resolveCallingMember falls back to a raw member UUID instead of failing closed, breaking credential-scoping by friendly name**

`getAgentOrFail(session.member_id)` returns `Agent | string`, where the string IS the error message (`Member "xyz" not found.`) -- it never throws. Line 97's `typeof agent === 'string' ? session.member_id : agent.friendlyName` treats that failure by falling back to the raw `session.member_id` UUID as the calling-member identity, rather than the synthetic `session:${sessionId}` identity already used one branch above (line 95) for the analogous 'session not found' case.

Failure scenario: a session resolves but its backing agent has since been removed/renamed in the registry (stale session, member deregistered mid-session). `resolveSecret`/`credentialResolve` then scopes the credential lookup against a raw UUID that will never match a friendly-name `allowedMembers` list (per `credential_store_set`'s docs, allowedMembers is a list of friendly names) -- a legitimate member gets a confusing 'denied' error referencing an internal UUID instead of a clear identity-resolution failure, and the doc comment's own fail-closed contract (lines 76-89) isn't followed consistently between the two fallback branches.

Suggest returning the same `session:${sessionId}` synthetic identity here for consistency with line 95, or explicitly surfacing the getAgentOrFail failure.

src/tools/send-email.ts
```diff
@@ -0,0 +1,149 @@
...
+}
+
+function buildProvider(input: SendEmailInput, callingMember: string): EmailProvider {
+  if (input.provider === 'smtp') {
+    const pass = resolveSecret('smtp_password', callingMember);
+    if (!pass) {
```
kumaakh: **Falsy check misreports a stored-but-empty credential as 'not found'**

`resolveSecret` returns `resolved.plaintext`, which can legitimately be an empty string if a caller stored an empty-string secret via `credential_store_set`. `if (!pass)` (and the identical `if (!apiKey)` at line 116) treats `''` the same as `undefined`, throwing 'SMTP password not found' even though the credential exists and was resolved.

Failure scenario: an operator runs `credential_store_set(name: "smtp_password", value: "")` (e.g. to intentionally clear/rotate a credential, or a bug upstream stores an empty string) -- `send_email` then reports a misleading 'not found. Store it with credential_store_set' error instead of the real 'credential is empty' state.

Use `pass === undefined` (or check `resolveSecret`'s return type more precisely) instead of the truthiness check.

src/tools/send-email.ts
```diff
@@ -0,0 +1,149 @@
...
+function resolveCallingMember(extra?: { sessionId?: string }): string {
+  const sessionId = extra?.sessionId;
+  if (!sessionId) return '*';
+  const session = sessionRegistry.findBySessionId(sessionId);
+  if (!session) return `session:${sessionId}`;
+  const agent = getAgentOrFail(session.member_id);
+  return typeof agent === 'string' ? session.member_id : agent.friendlyName;
+}
+
+function buildProvider(input: SendEmailInput, callingMember: string): EmailProvider {
```
kumaakh: **Provider selection is a hand-written if/else duplicating the resolve-secret-or-throw pattern per branch, not a factory over the EmailProvider interface**

Each branch of `buildProvider` independently resolves a hardcoded credential name, builds a custom not-found error string, and constructs a concrete provider class -- even though `src/providers/email/provider.ts` already defines an `EmailProvider` interface (and an unused `EmailConfig`/`EmailProviderName`) that could drive a table/factory instead.

Failure scenario: adding a third provider (SES, Mailgun) means growing this if/else and re-deriving the same 'resolve secret, throw formatted message, build config' boilerplate a third time, with nothing enforcing that new providers follow the same credential-resolution/error-message contract -- easy for a future provider to diverge in error wording or scoping behavior.

Consider a `{name, credentialName, build}` provider registry so adding a provider is a data addition, not new control flow.

src/providers/email/provider.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export interface EmailConfig {
+  provider: 'sendgrid' | 'smtp';
+  from: string;
+  sendgrid?: SendgridConfig;
+  smtp?: SmtpConfig;
+}
+
+export type EmailProviderName = 'sendgrid' | 'smtp';
```
kumaakh, line 52: **EmailConfig and EmailProviderName are dead exports**

Both are exported here and re-exported from `index.ts`, but nothing in this PR constructs or consumes an `EmailConfig` -- `buildProvider` in `send-email.ts` builds `SendgridConfig`/`SmtpConfig` object literals directly, and `input.provider` is typed via the zod enum, not `EmailProviderName`.

Cost: this looks like the intended config abstraction that provider selection should route through, but it's bypassed by the if/else in `buildProvider` -- a maintainer editing `EmailConfig` will reasonably assume it drives provider construction and be surprised it doesn't.

Delete both, or wire `buildProvider` through `EmailConfig` so the type actually matches how providers get built.

src/tools/send-email.ts
```diff
@@ -0,0 +1,149 @@
...
+import { SendGridProvider } from '../providers/email/sendgrid.js';
+import { SmtpProvider } from '../providers/email/smtp.js';
+import type { EmailMessage, EmailProvider } from '../providers/email/provider.js';
+import { sessionRegistry } from '../services/session-registry.js';
+import { getAgentOrFail } from '../utils/agent-helpers.js';
+import { logLine } from '../utils/log-helpers.js';
+
+// Basic RFC-5322-ish email format check -- not exhaustive, just catches
+// obviously malformed input before we attempt a network call.
+const EMAIL_RE = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
```
kumaakh: **Hand-rolled email regex duplicates zod's built-in .email() validator**

`EMAIL_RE = /^[^\s@]+@[^\s@]+\.[^\s@]+$/` reimplements address-format checking that `z.string().email()` already provides, and zod is already the schema library used for every field in this file.

Cost: this file's definition of 'valid email' can silently drift from what `z.string().email()` would accept/reject elsewhere in the codebase (e.g. plus-addressing edge cases, IDN), and any future fix to email-format validation has to be hand-applied here instead of picked up from a zod version bump.

Suggest `z.string().email()` (via `.refine`/`.safeParse` per address, since `to`/`cc`/`bcc` are arrays) instead of the custom regex.

src/tools/send-email.ts
```diff
@@ -0,0 +1,149 @@
...
+ * Fail-closed rule: only a stdio caller (no sessionId at all -- the fleet
+ * operator's own MCP connection) gets the '*' operator scope. An HTTP session
+ * that is NOT a registered member session (http-transport admits
+ * unauthenticated connections, and session churn can leave a sid
+ * unresolvable) gets a synthetic non-member identity instead of '*': it can
+ * still use credentials whose allowedMembers is '*' (the default -- this is
+ * what workflow clients rely on), but member-scoped credentials are denied
+ * rather than silently bypassed.
+ */
+function resolveCallingMember(extra?: { sessionId?: string }): string {
```
kumaakh: **resolveCallingMember duplicates the session -> member identity lookup already implemented in report-status.ts**

`sessionRegistry.findBySessionId(sessionId)` -> `session.member_id` -> agent lookup is the same sequence `src/tools/report-status.ts` already performs for the same purpose (mapping an MCP session to a caller identity), but as an independent, slightly-diverging implementation (this one additionally calls `getAgentOrFail` for a friendly name).

Cost: two independently-maintained implementations of 'map session -> member identity' now exist. A future fix to session-resolution semantics (e.g. tightening the fail-closed behavior for unresolvable sessions, per this function's own doc comment) has to be found and applied in both files, and they can silently diverge -- which is exactly the failure mode line 96-97 in this same function is already showing.

Suggest extracting a shared helper (e.g. into `src/utils/agent-helpers.ts`, which already hosts `getAgentOrFail`) that both tools call.

src/providers/email/smtp.ts
```diff
@@ -0,0 +1,375 @@
...
+}
+
+function dotStuff(message: string): string {
+  return message.split('\r\n').map(line => (line.startsWith('.') ? '.' + line : line)).join('\r\n');
+}
+
+export class SmtpProvider implements EmailProvider {
+  public readonly name = 'smtp';
+
+  constructor(private readonly config: SmtpConfig) {}
```
kumaakh: **Plaintext credential is retained on the provider instance for its whole lifetime, past the point it's needed**

`SmtpProvider` stores the resolved `SmtpConfig` (including `auth.pass`, the decrypted secret) as `private readonly config`. The plaintext password is only actually used once, in the AUTH LOGIN exchange (lines 339-346), but it stays reachable on the heap via `this.config` for the rest of the object's lifetime -- and `buildProvider`/`sendEmail` in `send-email.ts` keep the `provider` reference alive until `sendEmail` returns.

Cost: the decrypted credential lingers in memory longer than necessary (a live, reachable object property rather than a local that goes out of scope after use), widening the exposure window for anything that can inspect the process heap (core dump, debugger attach, memory-disclosure bug elsewhere in the process).

Consider passing `auth.pass` as a call argument to `send()` instead of storing it on `this.config`, or clearing it from the config object immediately after the AUTH step completes.

src/providers/email/smtp.ts
```diff
@@ -0,0 +1,402 @@
...
+    });
+  }
+
+  async command(cmd: string): Promise<SmtpResponse> {
+    const pending = this.waitForResponse();
+    this.socket.write(cmd + '\r\n');
+    return pending;
+  }
+
+  swapSocket(socket: tls.TLSSocket): void {
```
kumaakh, line 141: **STARTTLS response injection (CVE-2011-0411 class).** `swapSocket()` removes the plaintext socket's listeners and reattaches to the TLS socket, but never resets `this.buffer` (line 41) or `this.pending` (line 44). Between `STARTTLS` succeeding and `swapSocket()` running at line 361, the old `onData` handler (still attached, line 58) is live on the raw socket while `tls.connect({socket}, ...)` (in `starttls()`, line 180) simultaneously consumes bytes from the same underlying socket. A MITM on the plaintext leg can inject a complete SMTP-response-shaped line during this window; since no waiter is pending at that point, it lands in `this.pending` (line 107) and is never cleared by `swapSocket()`. The next `conn.command('EHLO localhost')` after the TLS upgrade then immediately `shift()`s that attacker-forged pre-TLS response instead of waiting for the genuine encrypted reply, desyncing the response stream and letting an attacker forge post-STARTTLS SMTP responses (e.g. fake AUTH/capability results). Fix: clear `buffer`/`pending` in `swapSocket()`, and detach the old socket's listeners before starting the TLS handshake rather than after it resolves.

src/providers/email/index.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/**
+ * Resolve a secret from the credential store, enforcing member scoping.
+ * `callingMember` must be the genuine caller identity: a member friendly name
+ * for member sessions, or '*' only for the fleet operator (orchestrator).
+ * Throws when the credential exists but is denied to the caller or expired,
+ * so those cases surface their real reason instead of "not found".
+ */
+export function resolveSecret(credentialName: string, callingMember: string): string | undefined {
```
kumaakh: **network_policy is never enforced for email credentials.** `resolveSecret()` only checks `'denied'`/`'expired'` on the resolved credential and returns `resolved.plaintext` -- it never inspects `resolved.meta.network_policy`. The shared `credentialResolve()` in `src/services/credential-store.ts` doesn't enforce this field either; it's returned as plain data for the caller to gate. `src/tools/execute-command.ts` (lines ~190-205) does enforce it (blocks `deny`, requires OOB confirmation for `confirm`) before letting a credential touch the network. `send_email`'s entire purpose is network egress with a stored secret, but this path silently skips that gate: an operator who sets `network_policy=deny` (or `confirm`) on `smtp_password`/`sendgrid_api_key` to block/gate its use over the network can still have it used to authenticate an outbound SMTP/SendGrid send. Suggest enforcing `network_policy` here the same way `execute-command.ts` does, or better, move the check into `credentialResolve()` itself so every caller gets it for free.

src/providers/email/smtp.ts
```diff
@@ -0,0 +1,402 @@
...
+    );
+  } else {
+    parts.push(`--${boundary}\r\nContent-Type: text/plain; charset=utf-8\r\n\r\n${msg.body}`);
+  }
+
+  for (const att of msg.attachments ?? []) {
+    assertHeaderSafe('attachment filename', att.filename);
+    if (att.contentType) assertHeaderSafe('attachment contentType', att.contentType);
+    parts.push(
+      `--${boundary}\r\nContent-Type: ${att.contentType ?? 'application/octet-stream'}; name="${att.filename}"\r\n` +
```
kumaakh, line 286: **Unescaped double quotes in MIME attachment headers.** `att.filename` (caller-supplied, `z.string().min(1)` with no quote restriction) is spliced directly into `name="${att.filename}"` and `filename="${att.filename}"`. `assertHeaderSafe()` only rejects CR/LF, not `"`. A filename like `report".txt; name="evil` breaks out of the quoted MIME parameter, producing a malformed/ambiguous header that receiving MIME parsers can mis-tokenize (e.g. truncate the visible filename or inject a bogus extra parameter). Suggest escaping embedded `"` (e.g. `replace(/"/g, '\\"')`) or rejecting filenames containing `"` in `assertHeaderSafe`/the attachment schema.

src/tools/send-email.ts
```diff
@@ -0,0 +1,148 @@
...
+  content: z.string().min(1).describe('Base64-encoded attachment content'),
+  contentType: z.string().optional(),
+});
+
+export const sendEmailSchema = z.object({
+  provider: z.enum(['sendgrid', 'smtp']).default('sendgrid').describe('Email provider to use'),
+  from: emailAddress.describe('Sender email address'),
+
+  host: z.string().optional().describe('SMTP server hostname (required for smtp provider)'),
+  port: z.number().optional().default(587).describe('SMTP server port'),
```
kumaakh: **`port` default (587) is independent of `secure`.** `port` defaults to 587 (STARTTLS) and `secure` defaults to `false` with no cross-field logic tying them together. A caller passing `{ provider: 'smtp', secure: true, host, user, from, to, subject, body }` without an explicit `port` gets `port=587` while `secure: true` makes `connectSocket` call `tls.connect({host, port})` immediately (implicit TLS) -- sending a TLS ClientHello to a server on 587 that expects plaintext SMTP first. This hangs until the connect/handshake timeout instead of failing fast with a clear error. Suggest defaulting `port` to 465 when `secure===true` and no explicit port is given (e.g. via `.superRefine`/a preprocessing step), or documenting the required pairing and validating it explicitly.

src/providers/email/smtp.ts
```diff
@@ -0,0 +1,402 @@
...
+  return words.join('\r\n ');
+}
+
+/** Wrap base64 content at 76 chars per line (RFC 2045). */
+function wrapBase64(content: string): string {
+  const compact = content.replace(/\s+/g, '');
+  return compact.replace(/(.{76})/g, '$1\r\n').replace(/\r\n$/, '');
+}
+
+function buildMimeMessage(msg: EmailMessage, from: string): string {
```
kumaakh, line 238: **Whole MIME message (incl. attachments) is built in memory with no size cap.** `buildMimeMessage()` returns the fully-assembled string (base64-inflated attachments included), which then goes through `normalizeCrlf()` and `dotStuff()` -- each a full split/map/join pass over the whole message -- before a single `conn.write()` call. Nothing streams to the socket until all of this completes, and there's no attachment/message size limit anywhere in the email path (grepped the repo -- only unrelated 64KB buffers elsewhere). For a message with multi-MB attachments this holds several full in-memory copies concurrently and is effectively uncapped, so a large or many-attachment `send_email` call is a memory/latency spike proportional to attachment size with no upper bound. Suggest at least a size limit on total attachment payload, and longer-term streaming the DATA payload to the socket instead of buffering the whole thing.

src/tools/send-email.ts
```diff
@@ -0,0 +1,148 @@
...
+ * tool registration only accepts a flat ZodRawShape (sendEmailSchema.shape),
+ * so the rule cannot live in the schema as a discriminated union.
+ */
+function validateInput(input: SendEmailInput): string[] {
+  const errors: string[] = [];
+  const toList = Array.isArray(input.to) ? input.to : [input.to];
+
+  if (toList.length === 0) errors.push(`'to' must include at least one recipient`);
+  for (const addr of toList) {
+    if (!emailAddress.safeParse(addr).success) errors.push(`Invalid 'to' address: ${addr}`);
```
kumaakh: **Redundant address validation.** `validateInput()` re-runs `emailAddress.safeParse()` on `to`/`cc`/`bcc`/`from` (lines 51,54,57,59) using the same `emailAddress` validator that `sendEmailSchema` already applies -- and `sendEmailSchema.shape` is registered directly with the MCP SDK (`tool-registry.ts`), which validates before the handler runs. By the time `validateInput` executes, every address has already passed this exact check once. The empty-`to`-list check, CRLF/subject check, and SMTP host/user cross-field check in the same function are NOT redundant (schema can't express them) -- only the four `emailAddress.safeParse` calls are pure duplication. Suggest dropping those four checks and keeping the rest.

src/tools/send-email.ts
```diff
@@ -0,0 +1,148 @@
...
+  for (const addr of input.cc ?? []) {
+    if (!emailAddress.safeParse(addr).success) errors.push(`Invalid 'cc' address: ${addr}`);
+  }
+  for (const addr of input.bcc ?? []) {
+    if (!emailAddress.safeParse(addr).success) errors.push(`Invalid 'bcc' address: ${addr}`);
+  }
+  if (!emailAddress.safeParse(input.from).success) {
+    errors.push(`Invalid 'from' address: ${input.from}`);
+  }
+  if (/[\r\n]/.test(input.subject)) {
```
kumaakh: **Duplicate, narrower CRLF/header-injection check.** This re-derives the exact same regex (`/[\r\n]/.test(...)`) as `assertHeaderSafe()` in `src/providers/email/smtp.ts`, but only applies it to `subject` here -- `from`/`to`/`cc`/`bcc` get format validation via zod but no CRLF check at this layer (they do get `assertHeaderSafe` one layer down in smtp.ts, but that's inconsistent coverage between the two layers). If `assertHeaderSafe` is later hardened (e.g. to also reject bare CR or NUL), this copy silently stays stale. Suggest exporting `assertHeaderSafe` (or moving it to a shared location) and calling it here instead of re-deriving the check.

src/providers/email/smtp.ts
```diff
@@ -0,0 +1,402 @@
...
+  /** Tear down the socket unconditionally (idempotent). Marks the
+   *  connection as deliberately closed BEFORE destroying, so the resulting
+   *  'close' event doesn't manufacture an "unexpected close" error. */
+  destroy(): void {
+    if (!this.failed) this.failed = new Error('SMTP connection already closed.');
+    this.socket.destroy();
+  }
+}
+
+function connectSocket(host: string, port: number, secure: boolean, timeoutMs: number): Promise<net.Socket | tls.TLSSocket> {
```
kumaakh: **`connectSocket()` and `starttls()` (line 180) are near-identical copy-paste.** Each wraps a socket connect/handshake in a `Promise` with: a `setTimeout` that destroys the socket and rejects with a timeout error, a success callback that clears the timer and resolves, and a `once('error')` handler that clears the timer and rejects. A shared `connectWithTimeout(connectFn, timeoutMs, timeoutMsg)` helper parameterized on the connect call and error message would eliminate the duplication and centralize a future fix to the timeout/cleanup logic (e.g. the STARTTLS injection issue flagged elsewhere in this review touches this exact cleanup ordering).

src/providers/email/smtp.ts
dsiddharth2: Fixed in c2c256d. prepareTlsUpgrade() now detaches plaintext listeners and clears buffer/pending before tls.connect({socket}). A forged reply in the STARTTLS window is dropped instead of being consumed as the first post-upgrade EHLO response. Test added that injects 550 FORGED after the STARTTLS 220 and confirms the session proceeds normally on the TLS socket.

src/providers/email/index.ts
dsiddharth2: Fixed in c2c256d. resolveSecret now enforces network_policy: deny fails closed, confirm uses the same OOB gate as execute_command. Unattended send_email needs network_policy=allow (the default). Test confirms network_policy=deny blocks the send before any provider is instantiated.

src/providers/email/smtp.ts
dsiddharth2: Fixed in c2c256d. Attachment filename and contentType reject double quotes, backslashes, CR, and LF -- in the Zod schema (quotedParam refine), in validateInput() via assertQuotedParamSafe(), and in SmtpProvider.send() before any SMTP I/O. Test confirms a filename like report".txt is rejected before credential lookup.

src/tools/send-email.ts
dsiddharth2: Fixed in c2c256d. Port is no longer hardcoded to 587. New smtpPort() helper returns 465 when secure is true and port is omitted, 587 otherwise. The schema description documents the default. Test confirms { secure: true } without port produces { port: 465 }.

src/providers/email/smtp.ts
dsiddharth2: Fixed in c2c256d. emailPayloadBytes() in headers.ts sums body + html + attachment content (base64 bytes, whitespace stripped), capped at MAX_EMAIL_PAYLOAD_BYTES (10MB). Checked in both validateInput() and SmtpProvider.send() via assertPayloadSize(). Test sends a 10MB+1 attachment and confirms rejection before credential lookup.

src/tools/send-email.ts
dsiddharth2: Fixed in c2c256d. The four emailAddress.safeParse() calls are gone from validateInput(). Format checks now live in sendEmailSchema (also applied to direct sendEmail() callers via safeParse at the top of the function). Cross-field checks (empty to, SMTP host/user, payload size) remain in validateInput().

src/tools/send-email.ts
dsiddharth2: Fixed in c2c256d. assertHeaderSafe() moved to src/providers/email/headers.ts -- single definition shared by both layers. validateInput() now uses headerError() (non-throwing wrapper) for from/to/cc/bcc/subject, so all header-bound values get the CR/LF check at validation time, not just subject.

src/providers/email/smtp.ts
dsiddharth2: Fixed in c2c256d. connectSocket() and starttls() now both delegate to a generic connectWithTimeout<T>() helper that owns the timeout/error/resolve pattern. Each caller is a thin wrapper passing its connectFn and timeout message.

src/tools/send-email.ts
dsiddharth2: Fixed in c2c256d. Extracted resolveSessionCaller() into src/utils/session-caller.ts -- both send_email and report_status now call the shared helper. The fail-closed semantics (unresolvable sessions get a synthetic session:<id> identity instead of a raw UUID) are defined once.

src/providers/email/smtp.ts
dsiddharth2: Fixed in c2c256d. SmtpProvider now stores auth.pass as a mutable private field, clears the caller's copy in the constructor (config.auth.pass = ''), then consumes its own copy into a local at the start of send() and sets this.pass = undefined. The plaintext is unreachable on the heap after AUTH completes.

src/providers/email/smtp.ts
dsiddharth2: Fixed in 219e479d. STARTTLS is now required when secure is false -- if the server does not advertise STARTTLS, the connection is aborted with an error before AUTH. Plaintext AUTH is never attempted.

src/tools/send-email.ts
dsiddharth2: Fixed in 219e479d. Extracted resolveSessionCaller() into src/utils/session-caller.ts. When getAgentOrFail returns a raw string (UUID) instead of an Agent, the caller gets a synthetic session:<id> identity instead of the UUID, so member-scoped credentials are denied rather than silently bypassed.

src/tools/send-email.ts
dsiddharth2: Fixed in 219e479d. resolveSecret now returns resolved.plaintext unconditionally (including empty string). The undefined return only happens when credentialResolve returns null (credential not found). An empty-string secret is a valid stored value, not a missing one.

src/tools/send-email.ts
dsiddharth2: Fixed in 219e479d. buildProvider now uses a PROVIDER_SPECS record keyed by provider name. Each spec declares its credentialName, missingMessage, and toConfig function. Adding a provider is a new record entry -- no if/else duplication.

src/providers/email/provider.ts
dsiddharth2: Fixed in 219e479d. EmailConfig and EmailProviderName are now consumed by the PROVIDER_SPECS record and createEmailProvider in src/providers/email/index.ts. They are live exports used by the provider factory, no longer dead.

src/tools/send-email.ts
dsiddharth2: Fixed in 219e479d. EMAIL_RE is gone. Both sendEmailSchema and validateInput use z.string().email() -- single definition of valid email format across the schema and runtime checks.

dsiddharth2:
Round-4 review comments addressed in c2c256d (replies inline on each thread). Also replied to the 2 remaining round-3 threads (session-caller dedup, credential retention) and the 6 original round-3 threads from 219e479d that were missing replies.

src/providers/email/index.ts
kumaakh: Confirmed fixed at head c2c256d. resolveSecret(name, callingMember) now requires an explicit caller, and send_email passes resolveSessionCaller(extra.sessionId).identity (src/utils/session-caller.ts); the operator * scope is used only for a sessionless/stdio caller. Resolving.

src/providers/email/smtp.ts
kumaakh: Confirmed fixed at head c2c256d. from/to/cc/bcc are z.string().email(), and headerError()/assertHeaderSafe() reject CR/LF in from/to/cc/bcc/subject in validateInput, then again in SmtpProvider.send() before any socket I/O. Resolving.

src/services/email/index.ts
kumaakh: Confirmed fixed at head c2c256d. src/services/email/ no longer exists on the branch, and there are no process.env reads anywhere in src/providers/email/ or src/tools/send-email.ts. Resolving.

packages/apra-fleet-client/src/client/api.mjs
kumaakh: Confirmed fixed at head c2c256d. SendEmailOptions (packages/apra-fleet-client/src/client/api.mjs:167-183) now documents provider, from (required), host, port, user, secure. Resolving.

src/tools/send-email.ts
kumaakh: Confirmed at head c2c256d. The smtp host/user rule now lives in exactly one runtime place (validateInput, src/tools/send-email.ts:85-88) with the MCP ZodRawShape constraint documented; the example workflow check is a local config.json fail-fast, not a third copy of the tool contract. Accepting the reasoning. Resolving.

src/tools/send-email.ts
kumaakh: Confirmed fixed at head c2c256d. sendEmail() runs the schema parse plus validateInput() (addresses, headers, host/user, payload size) before buildProvider() touches the credential store (src/tools/send-email.ts:169-186). Agreed on skipping the credential cache. Resolving.

src/providers/email/index.ts
kumaakh: Confirmed fixed at head c2c256d. resolveSecret(name, callingMember) now requires an explicit caller, and send_email passes resolveSessionCaller(extra.sessionId).identity (src/utils/session-caller.ts); the operator * scope is used only for a sessionless/stdio caller. Resolving.

src/providers/email/smtp.ts
kumaakh: Confirmed fixed at head c2c256d. from/to/cc/bcc are z.string().email(), and headerError()/assertHeaderSafe() reject CR/LF in from/to/cc/bcc/subject in validateInput, then again in SmtpProvider.send() before any socket I/O. Resolving.

src/services/email/index.ts
kumaakh: Confirmed fixed at head c2c256d. src/services/email/ no longer exists on the branch, and there are no process.env reads anywhere in src/providers/email/ or src/tools/send-email.ts. Resolving.

packages/apra-fleet-client/src/client/api.mjs
kumaakh: Confirmed fixed at head c2c256d. SendEmailOptions (packages/apra-fleet-client/src/client/api.mjs:167-183) now documents provider, from (required), host, port, user, and secure. Resolving.

src/tools/send-email.ts
kumaakh: Confirmed at head c2c256d. The smtp host/user rule now lives in exactly one runtime place (validateInput, src/tools/send-email.ts:85-88) with the MCP ZodRawShape constraint documented; the example workflow check is a local config.json fail-fast, not a third copy of the tool contract. Accepting the reasoning. Resolving.

src/tools/send-email.ts
kumaakh: Confirmed fixed at head c2c256d. sendEmail() runs the schema parse plus validateInput() (addresses, headers, host/user, payload size) before buildProvider() touches the credential store (src/tools/send-email.ts:169-186). Agreed on skipping the credential cache. Resolving.

examples/workflows/email-notify/main.mjs
kumaakh: Confirmed fixed at head c2c256d. The commented execute_command secret-piping block is gone; docs/email-workflow-guide.md:137-147 shows a direct CI step piping from the vault CLI over stdin, with an explicit warning against routing secrets through execute_command. Resolving.

src/providers/email/smtp.ts
kumaakh: Confirmed fixed at head c2c256d. connectWithTimeout() covers both the connect and the STARTTLS handshake, and waitForResponse() arms a per-response timer that destroys the socket on expiry (src/providers/email/smtp.ts:103-120, 170-207). Resolving.

examples/workflows/hello-world/main.mjs
kumaakh: Confirmed fixed at head c2c256d. examples/workflows/hello-world/main.mjs has no email phase on the branch; it is back to the hermetic version. Resolving.

docs/email-workflow-guide.md
kumaakh: Confirmed fixed at head c2c256d. docs/email-workflow-guide.md:43-50 imports and uses parseToolJson() from @apralabs/apra-fleet-client. Resolving.

README.md
kumaakh: Confirmed fixed at head c2c256d. README.md:289-296 now uses parseToolJson() with connectFleet(). Resolving.

src/providers/email/smtp.ts
kumaakh: Confirmed fixed at head c2c256d. onData() loops extracting every complete response, deliver() parks a response in this.pending when no waiter is queued, and waitForResponse() drains pending first (src/providers/email/smtp.ts:65-120). Resolving.

src/providers/email/smtp.ts
kumaakh: Confirmed fixed at head c2c256d. src/providers/email/smtp.ts:369-374 aborts when secure is false and the server does not advertise STARTTLS, so AUTH LOGIN is never reached over plaintext. Resolving.

src/tools/send-email.ts
kumaakh: Confirmed fixed at head c2c256d. resolveSessionCaller() returns the synthetic session:id identity when getAgentOrFail returns a string (src/utils/session-caller.ts:36-39); a raw member UUID is never used as the scoping identity. Resolving.

src/tools/send-email.ts
kumaakh: Confirmed fixed at head c2c256d. resolveSecret returns resolved.plaintext unconditionally, and buildProvider distinguishes undefined (not found) from an empty string (stored-but-empty) with separate messages (src/tools/send-email.ts:151-156). Resolving.

src/tools/send-email.ts
kumaakh: Confirmed fixed at head c2c256d. PROVIDER_SPECS (src/tools/send-email.ts:119-158) drives credential name, error text, and config construction; createEmailProvider(EmailConfig) builds the instance. Resolving.

src/providers/email/provider.ts
kumaakh: Confirmed fixed at head c2c256d. EmailConfig and EmailProviderName are now live: EmailProviderName keys PROVIDER_SPECS and EmailConfig is the return type of every toConfig and the input to createEmailProvider. Resolving.

src/tools/send-email.ts
kumaakh: Confirmed fixed at head c2c256d. EMAIL_RE is gone; const emailAddress = z.string().email() is the single definition used across the schema. Resolving.

src/tools/send-email.ts
kumaakh: Confirmed fixed at head c2c256d. Both src/tools/send-email.ts and src/tools/report-status.ts import resolveSessionCaller from src/utils/session-caller.ts. Resolving.

src/providers/email/smtp.ts
kumaakh: Confirmed fixed at head c2c256d. SmtpProvider holds the password in a mutable private field, blanks config.auth.pass in the constructor, and moves it into a local while setting this.pass = undefined at the top of send() (src/providers/email/smtp.ts:320-338). Resolving.

src/providers/email/smtp.ts
kumaakh: Confirmed fixed at head c2c256d. prepareTlsUpgrade() removes the data/error/close listeners and clears both buffer and pending before tls.connect (src/providers/email/smtp.ts:134-155, 377-381), so a reply injected in the STARTTLS window is discarded. Resolving.

src/providers/email/index.ts
kumaakh: Confirmed fixed at head c2c256d. resolveSecret() now reads resolved.meta.network_policy: deny throws before any provider is constructed, confirm goes through the same collectOobConfirm gate as execute_command (src/providers/email/index.ts:50-67). Resolving.

src/providers/email/smtp.ts
kumaakh: Confirmed fixed at head c2c256d. Attachment filename and contentType reject double quotes, backslashes, CR and LF at three layers: the quotedParam zod refine, assertQuotedParamSafe in validateInput, and again in SmtpProvider.send() / buildMimeMessage(). Resolving.

src/tools/send-email.ts
kumaakh: Confirmed fixed at head c2c256d. smtpPort() (src/tools/send-email.ts:114-117) returns 465 when secure is true and port is omitted, 587 otherwise; the schema description documents it. Resolving.

src/providers/email/smtp.ts
kumaakh: Confirmed fixed at head c2c256d. emailPayloadBytes() and MAX_EMAIL_PAYLOAD_BYTES (10MB) in src/providers/email/headers.ts are enforced in validateInput() before credential resolution, and again via assertPayloadSize() in SmtpProvider.send(). Resolving.

src/tools/send-email.ts
kumaakh: Confirmed fixed at head c2c256d. The four redundant emailAddress.safeParse() calls are gone; validateInput() keeps only the checks the flat ZodRawShape cannot express. Resolving.

src/tools/send-email.ts
kumaakh: Confirmed fixed at head c2c256d. assertHeaderSafe() now lives in src/providers/email/headers.ts and is the single definition used by both layers; validateInput() applies headerError() to from/to/cc/bcc/subject, not just subject. Resolving.

src/providers/email/smtp.ts
kumaakh: Confirmed fixed at head c2c256d. connectSocket() and starttls() are both thin wrappers over the generic connectWithTimeout helper (src/providers/email/smtp.ts:170-207). Resolving.

kumaakh:
## Convergence review (head c2c256d)

Reconciled all 28 open review threads against the current code rather than the diff at review time: **28/28 verified fixed in code and resolved**, each with a confirming reply pointing at the current file/lines. Two were closed as accepted-reasoning rather than code changes (the discriminatedUnion request, blocked by the MCP SDK ZodRawShape contract, and the deliberate no-credential-cache decision) -- both are documented in code.

Independent security pass over the credential-handling core (not relying on the prior rounds):

- **Secret leakage**: no path leaks a secret. SMTP errors carry only server response lines; the password is only ever sent base64-encoded and never interpolated into an error. SendGrid errors carry the response body, not the Authorization header. `logLine` only sees those same messages.
- **SMTP/MIME injection**: `from`/`to`/`cc`/`bcc` are `z.string().email()` plus `assertHeaderSafe` at both layers; `subject` is CRLF-checked and RFC 2047 encoded when non-ASCII; attachment `filename`/`contentType` reject `"`/`\`/CR/LF; `user` is base64-encoded before transmission; `host`/`port` never reach a shell. Body/html go through `normalizeCrlf` + `dotStuff`, closing the SMTP-smuggling path, and the MIME boundary is 12 random bytes.
- **Credential scoping**: `resolveSecret` takes an explicit caller from `resolveSessionCaller`, which fails closed to `session:<id>` for unresolvable sessions. `network_policy` is now enforced (deny/confirm) before any egress.
- **Zero-env-var**: verified true -- no `process.env` reads anywhere under `src/providers/email/` or `src/tools/send-email.ts`; the old `src/services/email/` copy with the env fallback is deleted.
- TLS certificate validation is correct in both modes, including the STARTTLS leg (`tls.connect({socket, host})` keeps hostname verification).

No P1 or P2 issues remain. Only P3 nits (long ASCII `Subject:` is not folded at 998 chars; `EHLO localhost` is hardcoded; the OOB confirm prompt is labelled `execute_command`) -- none worth another round.

**Recommendation: merge.**

PR #422, opened 2026-08-18
fix(providers): map unattended=auto and dangerous to provider-native flags for opencode and agy

kumaakh:
## Problem

The `unattended` permission mode handling for opencode and agy providers was incorrect:

- **opencode**: `unattended='dangerous'` passed `--dangerously-skip-permissions` which doesn't exist on opencode. `unattended='auto'` was not handled at all.
- **agy**: `unattended='auto'` was not handled, leaving the member to run without any permission shortcut.

## Solution

Both `auto` and `dangerous` now map to the provider-native permission flag:

- **opencode**: Both map to `--auto` (opencode has no `--dangerously-skip-permissions`)
- **agy**: Both map to `--dangerously-skip-permissions` (agy has no `--auto`)

Fallback warnings are logged when the requested mode is not natively supported.

This resolution now lives behind a single provider interface method, `resolvePermissionFlag(unattended)`, implemented per-provider (claude, codex, copilot, gemini, opencode, agy, none). Both dispatch paths -- the POSIX `buildPromptCommand()` (via `os/linux.ts`) and Windows (`os/windows.ts`) -- call this one method instead of each re-deriving the auto/dangerous/default branching independently. Previously `os/windows.ts` had its own duplicate branch that called the lower-level flag methods directly and did not replicate opencode/agy's fallback logic, so on Windows dispatch `unattended='auto'` silently added NO flag at all for opencode and agy -- the actual mechanism behind the stall/hang this PR fixes on non-Windows. Windows dispatch now gets the identical fallback behavior as POSIX.

## Unrelated-looking fix included here

`opencode.ts`'s `resolveSessionLogDir()` changes the session-log path from `.local/share/opencode/storage/chats` to `.local/share/opencode/log`. This is not scope creep: OpenCode's stall detector polls this path to confirm the member process is still writing, and it was pointed at the wrong directory (opencode does not write transcripts under `storage/chats`), so stall detection was silently non-functional for OpenCode members regardless of this PR's permission-mode fix. Both fixes are needed together for OpenCode dispatch to work reliably unattended.

## Safety Note

Neither opencode's `--auto` nor agy's `--dangerously-skip-permissions` have a classifier safety mechanism like Claude's `--permission-mode auto`. Ensure deny rules are properly configured for unattended members.

## Tests

- `tests/unattended-mode.test.ts`: 3 new cases covering the new opencode/agy behavior.
- `tests/windows-pid-wrap.test.ts`: 4 new cases pinning the Windows dispatch path to the same per-provider fallback semantics as POSIX (AGY auto/dangerous -> `--dangerously-skip-permissions`; OpenCode auto/dangerous -> `--auto`), regression-guarding the silent-no-flag bug described above.

### DIFF
--- a/src/os/windows.ts
+++ b/src/os/windows.ts
@@ -168,19 +168,13 @@ export class WindowsCommands implements OsCommands {
       const rf = provider.resumeFlag(sessionId, resuming);
       if (rf) argList += ` ${rf}`;
     }
-    if (unattended === 'auto') {
-      const autoFlag = provider.permissionModeAutoFlag();
-      if (autoFlag) argList += ` ${autoFlag}`;
-    } else if (unattended === 'dangerous') {
-      argList += ` ${provider.skipPermissionsFlag()}`;
-    } else {
-      // apra-fleet-eft.65.1: interactive-session parity for the work folder --
-      // grant Edit/Write of a brand-new file in the dispatched agent's own work
-      // folder without the broad --dangerously-skip-permissions bypass. Providers
-      // without such a surgical flag omit the method, leaving behavior unchanged.
-      const editFlag = provider.workspaceEditPermissionFlag?.();
-      if (editFlag) argList += ` ${editFlag}`;
-    }
+    // Delegate unattended-mode flag resolution entirely to the provider --
+    // each provider's own auto/dangerous fallback and warning semantics (e.g.
+    // AGY has no true auto and falls back to its dangerous flag; OpenCode has
+    // no true dangerous and falls back to --auto) must not be re-derived here,
+    // or this path silently diverges from the POSIX buildPromptCommand() path.
+    const permFlag = provider.resolvePermissionFlag(unattended);
+    if (permFlag) argList += ` ${permFlag}`;
     if (model) {
       argList += ` ${provider.modelFlag(escapeWindowsArg(model))}`;
     }
--- a/src/providers/agy.ts
+++ b/src/providers/agy.ts
@@ -84,7 +84,10 @@ export class AgyProvider implements ProviderAdapter {
       }
     }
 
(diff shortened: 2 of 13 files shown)


kumaakh:
Windows CI failure on this PR's run (build-and-test windows-latest, job 95803070721) is a pre-existing, unrelated flake -- not caused by this PR's changes.

**What failed:** `packages/apra-fleet-se/test/supervisor-stop-restart-integration.test.mjs` -- subtests (1) and (2) PASS, then the whole file is reported FAIL with no attached error/stack (subtests (3)/(4) never even print). The full raw job log has no stack trace or STDERR for this file.

**Why it's unrelated to this PR:** this branch's own commits (80b9a557, 5d8693dc, 9ae5a567) only touch `src/os/windows.ts`, `src/providers/*.ts`, and `tests/windows-pid-wrap.test.ts` -- nothing under `packages/apra-fleet-se/src/supervisor/`. ubuntu-latest and macos-latest passed on the identical commit.

**Repro attempts (all negative):** ran the file in isolation and the full apra-fleet-se suite (matching CI's exact `--test-concurrency=8` invocation) three times in a row on a real Windows machine -- 2239/2239 passing every time. Checked the last 3 windows-latest failures on `main` -- all fail for unrelated reasons (a `bd bootstrap` dolt-clone race, unrelated vitest integ tests), none reproduce this signature.

Tracked separately as `apra-fleet-buow` (P1, full evidence + suspected root cause: this test spawns real detached child processes under `--test-concurrency=8` on a 2-vCPU windows-2022 runner, which is plausibly resource-constrained enough to crash the per-file test process outside any JS try/catch). Safe to disregard the red Windows check on this PR and re-run if needed.

PR #425, opened 2026-08-18
fix(pm): relocate misplaced repo-root skills/pm/ and fix SEA install crash

kumaakh:
## Summary

Three commits addressing the reported crash and its full root cause.

### 1. Immediate crash fix

A third party downloading `apra-fleet-installer-win-x64.exe` from the latest CI build on `main` hit:
```
[7/14] Installing PM skill...
[fleet:error] cli Install failed: Cannot find project root (version.json not found)
```
`install.ts`'s PM-skill step called `findProjectRoot()` unguarded, which throws when there's no repo checkout anywhere on disk (true for a standalone downloaded SEA binary).

### 2. Root cause: misplaced repo-root skills/pm/

PR #305 (94bc1368) added a *new* `skills/pm/` directory at the repo root without realizing the PM skill's real source of truth had already moved to `packages/apra-fleet-se/apra-pm/skills/pm/` in an earlier restructure (apra-pm is a first-party monorepo package, not vendored/submodule content -- see 633a6bab). `install.ts` was trying to overlay that orphaned directory onto the installed skill, which is what actually threw. Fixed by:
- Removing the dead overlay step from `install.ts` entirely.
- Relocating the 4 live skill docs (`index.md`, `kb-agent.md`, `kb-reconcile.md`, `kb-review.md`) as-is into the correct package path.
- Deleting the 5 `tpl-*.md` templates (they implement an old PLAN.md/progress.json execution model, superseded by the current beads-driven agent system).

### 3. Recovering the tpl-*.md content properly, and a permissions audit

The `tpl-*.md` templates had genuinely useful Knowledge Bank guidance not yet ported to the current agents. Rather than restore them wholesale (which would reintroduce a stale execution model and a direct-tool-call capture pattern the engine has since superseded), this commit:
- Merges the KB guidance (retrieve-first-then-read-source, dedup-before-capture, `kb_feedback`, `kb_stats`-driven model tiering) into `agents/doer.md`, `agents/reviewer.md`, `agents/planner.md`, reconciled against the real current architecture (`auto-sprint.js`'s engine-executed `kb_captures`/`kb_promotions` structured-output pipeline -- the role decides, the engine calls the tool).
- Adds a new `agents/kb-reconciler.md` for the one tpl-role that's still live (`/pm kb-reconcile` Step 4).
- Enforces (and audits for) a dependency-direction rule: agents must never reference the pm skill or a `/pm` command; only the skill may reference agents. Found and fixed one violation in `harvester.md`.
- Wires the MCP tools each role actually calls into the fleet permission profiles (`skills/fleet/profiles/base-dev.json`/`base-reviewer.json`, new `tag-kb-reconciler.json`) -- these had zero `mcp__` entries despite every role instructing their use. Uses the real `mcp__apra-fleet-member__<tool>` format (not `mcp__apra-fleet__<tool>`, which is the member's own disabled local server instance).
- Fixes the AGY and OpenCode provider permission transforms: AGY's `convertClaudeAllowToAgyPermissions` only recognized a fictional `Mcp(name)` token, not the real `mcp__<server>__<tool>` format -- fixed. OpenCode's `composePermissionConfig` genuinely has no per-tool/per-server MCP permission concept in its schema (confirmed against `docs/opencode-exploration.md`'s live investigation) -- documented in place rather than inventing one.
- Along the way, found and fixed a real drift bug: `auto-sprint.js`'s inline `vetKbWork()` (a hand-duplicated copy, required since Workflow-tool scripts can't import files) had fallen out of sync with the canonical `lib/vet-kb-work.mjs`, silently dropping every capture's `content` field.

## Test plan
- [x] `tests/install-sea-no-project-root.test.ts` reproduces the original third-party crash scenario -- confirmed it fails pre-fix, passes post-fix.
- [x] `npx tsc --noEmit` clean.
- [x] Full `npx vitest run`: 4079 passed; the only failures are 3 pre-existing, unrelated KB-anchor git-clone-timeout tests (present before this branch, untouched by this change).
- [x] `apra-pm` package's own `node --test`: 439/439 pass, including `agent-schema-validation.test.mjs` and `agent-contracts-kb-wiring.test.ts` (updated for the new `kb-reconciler` role and the reviewer capture-mechanism change).
- [x] New AGY integration test for the `mcp__<server>__<tool>` -> server-granular permission-rule collapse.


joiskash:
## Review of #425 -- findings and decisions

Reviewed all 5 commits against `origin/main` (`3d26a320`). The PR's own work is sound:
the SEA install crash fix, the relocation out of the orphaned repo-root `skills/pm/`, the
tpl-to-agent content recovery, the AGY privilege-escalation guard, and the `kb_promote`
`repo_path` fix all hold up. `e81280de`'s self-reversal on the `apra-fleet-member` server
is correct -- `registerMcpEndpoint`'s only caller is behind `interactiveBootstrapEnabled()`,
hardcoded `return false` (`src/tools/register-member.ts:111`), so that server exists for
nobody today.

Four gaps found, all doc/config level. Fixed on the branch in 085e2f7d:

**1. The three relocated commands were orphaned inside the skill.** `SKILL.md` has two
indexes -- the command table and the file index -- and neither listed `index.md`,
`kb-reconcile.md`, or `kb-review.md` (`grep -in "index\|reconcile\|kb-review" SKILL.md`
returned nothing). Relocation was right; wiring was missing. Added three rows and three
file-index entries.

**2. `index.md` claimed two automation triggers that do not exist.** It said indexing runs
"Automatically inside `/pm init`" and "Automatically at VERIFY checkpoints". `/pm init`
routes to `sprint.md` Setup, which is four steps and indexes nothing; and `gitnexus`
appears in exactly one file across the whole skill + agents tree -- `index.md` itself.
Both bullets deleted, leaving the three real triggers.

**3. `kb-reconcile.md` claimed a `/pm cleanup` post-merge hook that exists nowhere.**
`sprint.md`'s Completion checklist is 8 mandatory items, none of them kb-reconcile, and
SKILL.md never mentions it. Same defect class as the dangling `tpl-kb-agent.md` reference
`3ae2f308` deleted. Reworded to describe the standalone command it actually is. Whether
`/pm cleanup` *should* run it is a real question -- filed separately rather than decided
here.

**4. The new `mcp__apra-fleet__*` profile grants are inert.** Kept as-is (zero behavioral
risk) with a `_comment` in each of `base-dev.json`, `base-reviewer.json`, and
`tag-kb-reconciler.json` recording why, so the next reader does not conclude members have
KB access. Two independent reasons they grant nothing today: a plain remote member has no
apra-fleet MCP server registered at all (see the gating above), and a member that is
itself an apra-fleet install has the whole server disabled by `composePermissionConfig`
in that same `settings.local.json`. Roles reach the KB through the orchestrator instead --
`runner.js`'s `kbKnowledgeBlock` and `auto-sprint.js`'s `runKbWork`. Worth noting the AGY
side-effect: each entry now hits the new refusal branch (`src/providers/agy.ts:468-487`),
so an AGY member compose emits 9 warnings and writes 9 `custom` rules.

`tag-kb-reconciler.json` is separately unreachable: tag profiles apply to members via
`compose_permissions`, but `kb-reconciler` is never dispatched to a member -- it is a local
subagent of `/pm kb-reconcile`, and appears in neither `auto-sprint.js`'s `agentType`
dispatches nor `fleet-sprint` `contracts.mjs`'s 9-role `ROLES`.

### Out of scope for this PR -- filed as issues

These predate #425 (they arrived with #305) and are tracked separately so this PR stays a
cleanup PR:

- #429 -- `docs/code-intelligence-providers.md` disagrees with the code twice: on which
  provider is actually the default (install writes gitnexus, the `getProvider()` fallback and
  the doc both say codebase-memory, every runtime path is gitnexus-only), and on per-member
  routing (documented as "not yet wired"; `code-intelligence.ts:127-139` wires it).
- #430 -- nothing indexes a cold repo. `maybeScheduleReindex` is reachable only from
  `computeFreshnessNote`, which needs an existing `.gitnexus/meta.json`; auto-reindex
  refreshes, it never creates. `/pm index` is the only entry point.
- #431 -- `install` never prunes. `copyDirSync` only copies, so `tpl-*.md` and `kb-agent.md`
  still sit in `~/.claude/skills/pm/` on every machine that installed before this PR --
  including the dangling `tpl-kb-agent.md` instruction `3ae2f308` deleted precisely so it
  would not ship.
- #432 -- the unverified assumption below.
- #433 -- should `/pm cleanup` run kb-reconcile post-merge, as kb-reconcile.md claimed?

### One unverified assumption

I could not confirm that Claude Code honors `mcpServers.<name>.disabled` inside
`.claude/settings.local.json`. If it does not, a *local* member inherits the user-scope
apra-fleet server -- which would mean the loop-prevention does not work and those grants
are live for that case. It does not change the conclusion for remote members. Filed as #432, and linked
from the `_comment` in both base profiles so the claim is not read as settled.

PR #446, opened 2026-08-21
fix(fleet-sprint): redact secure tokens from untrusted prompt blocks

joiskash:
A knowledge-bank entry whose title merely named the braced secure token was primed into every agent prompt of a sprint. execute_prompt rejects the WHOLE dispatch on a match (src/tools/execute-prompt.ts:255,497), so every dispatch -- planner and plan-reviewer alike -- came back in 2-6ms with zero LLM calls, zero tokens and zero cost. The rejection string was then parsed as if it were the model's reply, and the sprint aborted at PLAN_REVIEW_DISPATCH_FAILED.

The guard is correct for caller-authored prompts, but injected content is not caller-authored: an entry that names the token carries no secret. Redact at wrapUntrustedBlock, the single choke point all six injection sites in runner.js already go through (KB primed entries, kb_list candidates, bead acceptance criteria, plan-reviewer.notes, reviewer.notes), so no future entry or bead text can reintroduce the failure.

Redaction keeps the token NAME -- that is the part that makes the entry useful -- and appends a one-line note so a reader knows the braces were removed and must be written back when the token is actually used.

Also rewords the .fleet/kb-canonical.json entry that triggered this, since the bible is re-imported on every prime.

Verified: the exact entry that killed the sprint no longer matches the guard pattern once wrapped, the token name survives, and the note is present. Suites green apart from contracts-schema-dist-staleness-guard, which fails identically on a clean tree (stale dist/, predates this change).

Claude-Session: https://claude.ai/code/session_01LmjDVeqSGwSnbuPRokLkMU

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Testing

- [ ] Ran `npm test` — all tests pass
- [ ] Ran `npm run build` — build succeeds
- [ ] Manually tested the affected functionality

## Checklist

- [ ] My code follows the existing style and patterns in this codebase
- [ ] I have updated documentation where necessary
- [ ] Breaking changes are noted above and in the commit message
- [ ] No new linting errors introduced

### DIFF
--- a/.fleet/kb-canonical.json
+++ b/.fleet/kb-canonical.json
@@ -58,8 +58,8 @@
     {
       "id": "16ba5328-63a9-4f98-8614-fa3137afe2f3",
       "type": "knowledge",
-      "title": "{{secure.NAME}} values arrive already shell-escaped, and the escaping is OS-branched",
-      "summary": "execute_command replaces each {{secure.NAME}} token with a value that has ALREADY been quoted for the target member's shell -- escapePowerShellArg for windows members, escapeShellArg otherwise, chosen from agentOs. Reference the token bare in a command string; wrapping it in your own quotes double-escapes the value and surfaces as a false 401 or invalid-token error.",
+      "title": "secure.NAME token values arrive already shell-escaped, and the escaping is OS-branched",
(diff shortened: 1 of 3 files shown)


kumaakh:
we should redact every secret from the KB (trusted or untrusted)

PR #450, opened 2026-08-22
feat(fleet-sprint): non-exclusive orchestrator role via Agent.unreservable

kumaakh:
## Summary
- Adds `Agent.unreservable`: a member can be flagged as shared-by-construction, so `reserve`/`release`/`force_release` become no-ops and the supervisor's member-overlap guard skips it -- letting more than one concurrent sprint use the same orchestrator member instead of exclusively claiming it (previously a 409 `member overlap` on the second sprint).
- Lands the safe, backward-compatible subset of `docs/design-orchestrator-worktree-model-v2.md` section 4 (eliminating the orchestrator's git surface): drops `orchestratorMember` from `branchEnsureMembers`, delegates the `deploy.md` permissions read and the three file-existence probes to the actual role member about to consume the file, and resolves the abort-path git operations to the harvester/doer member instead of the orchestrator.
- Mirrors the schema change in `packages/apra-fleet-client`.

## Deferred (filed as follow-ups, not in scope here)
- `apra-fleet-xd4b`: concurrency serialization for a shared orchestrator's beads clone -- pull-collapsing, local-write serialization, and the same for KB writes. This needs a keyed generalization of the existing global push-mutex (`dolt-mutex.mjs`), plus HTTP+MCP client mirrors, since sprints run as separate detached processes. Comparable in size to the existing push-mutex system; needs its own design/test rigor.
- `apra-fleet-8zr3`: static `repo_url` launch config replacing `git remote get-url origin`, removing the now-redundant defensive push, the supervisor auto-injecting `roleMap.orchestrator`, and then hard-failing (rather than silently falling back) on an unmapped orchestrator role. Confirmed empirically that closing the fallback before the auto-injection ships breaks ~139 existing tests that rely on the implicit fallback -- these must land in that order.

## Test plan
- [x] `npm run build` (tsc) -- clean
- [x] `packages/apra-fleet-se` full suite: 2372 pass, 0 fail
- [x] Manually reproduced the original bug (two sprints both wanting `roleMap.orchestrator: ["supervisor"]` -> 409) and confirmed the fix unblocks it once the member is flagged `unreservable`

### DIFF
--- a/packages/apra-fleet-client/src/client/api.mjs
+++ b/packages/apra-fleet-client/src/client/api.mjs
@@ -104,6 +104,7 @@
  * @property {string} [category] - Optional group label
  * @property {string[]} [tags] - Optional list of free-form labels
  * @property {"false" | "auto" | "dangerous"} [unattended] - Permission mode for unattended execution
+ * @property {boolean} [unreservable] - Mark this member as never exclusively reservable, so it can be shared by more than one sprint at once (e.g. fleet-sprint's shared "orchestrator" role)
  */
 
 /**
@@ -122,6 +123,7 @@
  * @property {string} [category] - Group label
  * @property {string[]} [tags] - Free-form labels
  * @property {"false" | "auto" | "dangerous"} [unattended] - Permission mode
+ * @property {boolean} [unreservable] - Mark/unmark this member as shared/never exclusively reservable
  */
 
 /**
--- a/packages/apra-fleet-se/bin/cli.mjs
+++ b/packages/apra-fleet-se/bin/cli.mjs
@@ -670,8 +670,27 @@ async function main() {
         }
         return res && res.content && res.content[0] ? res.content[0].text : '';
     };
+    // apra-fleet: exclude any member ALSO named in roleMap.orchestrator from
+    // this git-identity/git-remote probe -- a shared/unreservable orchestrator
+    // member (docs/design-orchestrator-worktree-model-v2.md) may have no real
+    // checkout at all, and `git rev-parse HEAD` on one hard-fails the launch
+    // (process.exit(1) just below) even though the orchestrator role itself
+    // never needs to pass this check. This only matters when an operator also
+    // lists that member in --members (redundant with roleMap.orchestrator,
+    // and not required); a normal launch that passes the shared orchestrator
+    // ONLY via roleMap.orchestrator was never affected.
+    const orchestratorRoleMapMembersForTopology = new Set(
+        (roleMap && Array.isArray(roleMap.orchestrator)) ? roleMap.orchestrator : []
+    );
+    // Degrade back to the unfiltered list if excluding the orchestrator would
+    // empty it out entirely (e.g. a single-member sprint that role-maps its
+    // one dispatch member as orchestrator too) -- checkMemberTopology refuses
+    // to start on an empty member list, and that degenerate case is exactly
+    // the topology a real (non-shared) dispatch member should still pass.
+    const topologyMembersFiltered = validMembers.filter((m) => !orchestratorRoleMapMembersForTopology.has(m));
+    const topologyMembers = topologyMembersFiltered.length > 0 ? topologyMembersFiltered : validMembers;
     const topology = await checkMemberTopology({
-        members: validMembers,
+        members: topologyMembers,
         mode: syncedMode ? 'synced' : 'legacy',
         getIdentity: (member) => runCommand('git rev-parse HEAD', member),
         getOriginUrl: (member) => runCommand('git remote get-url origin', member),
--- a/packages/apra-fleet-se/bin/cli.mjs
+++ b/packages/apra-fleet-se/bin/cli.mjs
@@ -603,11 +603,19 @@ async function main() {
 
     // 2. Validate members exist via the fleet's list_members tool.
     let validMembers = [];
+    // apra-fleet: names flagged unreservable in the SAME list_members read --
+    // hoisted out of this try block so the topology filter below (item 4) can
+    // key on the actual unreservable flag, not on roleMap.orchestrator
+    // membership (which would also match a real, git-having dispatch member
+    // that is ADDITIONALLY role-mapped as orchestrator, and wrongly skip its
+    // legitimate same-HEAD topology check).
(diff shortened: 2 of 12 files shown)


kumaakh:
## Update: fixed the blocking findings from an independent Opus review

The review (run against this branch's actual diff + a fresh test/build run) came back CHANGES NEEDED with 4 blocking items. All 4 are now fixed in commit 22bc39dd:

1. **`unmappedRoleFallbackPool` could re-select the orchestrator for other roles** when every physical member was already a roleMap specialist -- silently undoing the branchEnsureMembers/publishGitMember/probeFileExists fixes. Fixed: roleMap.orchestrator members are now excluded from that pool entirely.
2. **`checkMemberTopology` (bin/cli.mjs) hard-fails on a git-less orchestrator listed in `--members`.** Fixed: any member also named in `roleMap.orchestrator` is excluded from that probe (degrading back to the unfiltered list only if that would empty it out).
3. **Zero test coverage for `unreservable`.** Added: member-reservation.ts suite (all 3 actions no-op, pre-existing `reservedBy` untouched), a supervisor-api.test.mjs case proving an unreservable orchestrator overlapping an active reservation does NOT 409, and a `defaultMemberOverlapGuard` case proving a stale server-side `reservedBy` on an unreservable member is skipped.
4. **`unreservable: true` had no pairing constraint with `llm_provider`,** letting an ordinary dispatch member (or a self-escalating one via its own `update_member` call) be flagged unreservable, which could interleave two sprints' dispatches into one working tree. Fixed: `register_member`/`update_member` now reject `unreservable: true` unless `llm_provider` is (or is being set to) `'none'`.

Also fixed two non-blocking items from the same review: `unreservableMemberNames` now logs a `listMembers()` failure instead of silently swallowing it, and `launch()` no longer opens two separate MCP transports per request for what is logically one `listMembers()` read.

**Known caveat from the review, not fixed here (tracked in `apra-fleet-xd4b`):** two sprints sharing one orchestrator member race on that member's beads clone at runtime -- not a data-integrity risk (Dolt push is already globally mutexed and versioned), but several `bd`/Dolt call sites use `fatal: true` or no `failSoft`, so a concurrent pull/write collision is more likely to abort a sprint than it would be with one orchestrator per sprint. This is the deferred concurrency-serialization work, not a regression from this branch.

Full test suite: 2374/2374 pass. Build clean. Re-review requested on the updated diff.

kumaakh:
## Round-2 fixes applied + final targeted re-verification: APPROVED

Independent background review ran two rounds against this branch's two acceptance criteria:
- (a) orchestrator can be used by multiple sprints created by the same supervisor
- (b) orchestrator role does not need a code checkout, though it will know the git remote and will have beads accessible to it

**Round 1** found 2 blocking correctness issues + 2 blocking test-coverage gaps, all fixed and re-verified closed:
- `unmappedRoleFallbackPool` degenerate fallback re-exposed the shared orchestrator to other roles when every physical member was already a roleMap specialist -- fixed by excluding `roleMap.orchestrator` members from the pool and its fallback.
- `checkMemberTopology` (bin/cli.mjs) hard-failed on a git-less shared orchestrator listed in `--members` -- fixed (see round 2 below for the corrected version).
- Zero test coverage for the `unreservable` no-op semantics -- added 4 tests in `tests/member-reservation.test.ts` + 2 in `packages/apra-fleet-se/test/supervisor-api.test.mjs`.
- `unreservable`/`llm_provider` pairing unenforced -- added enforcement in `register-member.ts` and `update-member.ts` (register-time hard block; update-time is airtight since `updateMemberSchema`'s `llm_provider` enum has no `'none'` option, so a member can never be moved into `'none'` via update).

**Round 2** found my round-1 topology-filter fix itself introduced a regression (it also skipped the same-HEAD check for a *legitimate* git-having member that happened to be dual-role-mapped as orchestrator), plus flagged the round-1 `listMembers()` single-flight dedup as ineffective (calls are sequential, not concurrent, so it never actually deduped). Both corrected in `470dcd37`:
- `bin/cli.mjs`: topology filter now keys off the member's `unreservable` flag (from the existing `listMembers()` fetch) instead of `roleMap.orchestrator` membership -- dual-role members now correctly get their real topology check.
- `src/supervisor/api.mjs`: removed the ineffective single-flight wrapper; `launch()` now does exactly one `listMembers()` fetch and passes it through as `membersList` to `defaultMemberOverlapGuard`, which uses it directly instead of re-fetching.
- Added a doc note (`fleet-sprint/docs/README.md`) that a shared/unreservable orchestrator must be passed via `roleMap.orchestrator` only, never `--members`.

**Final targeted re-verification** (separate agent instance, this branch, commit `470dcd37`) confirmed both fixes correct with no new regression, full test suite green (2374 pass / 0 fail, unchanged from baseline), clean build, and both acceptance criteria independently traced through the code and confirmed met. Verdict: **APPROVED**.

### Explicitly deferred (by design, not oversight)
- `apra-fleet-9nz7` (new, P3): `reReserveForResume`/`resyncReacquiredMember` run hard git steps against every `--members` entry on sprint resume; a *misconfigured* shared orchestrator additionally listed in `--members` (instead of only `roleMap.orchestrator`, the documented convention) would get wedged on resume. Does not affect correct usage.
- `apra-fleet-8zr3` (P2): remaining git-removal work from `docs/design-orchestrator-worktree-model-v2.md` sections 4.2/4.3/6.2/6.4 -- static repo_url config, dropping the defensive Publish-PR push, supervisor auto-injection of `roleMap.orchestrator`, and (only after auto-injection lands) making an unmapped orchestrator role a hard launch-time failure instead of the current silent fallback. Sequencing constraint documented in the bead and in-code (`TODO(orchestrator-hard-fail)`): the hard-fail change cannot land before auto-injection without breaking ~139 existing tests that rely on the implicit fallback.
- `apra-fleet-xd4b` (P1): keyed beads-mutex for serialized concurrent access (pull-collapsing, local-write serialization, extended to KB) -- explicitly scoped as a separate follow-up branch/PR per prior discussion, not a blocker for this PR. Not a data-integrity risk today (Dolt push is globally mutexed and versioned) but can increase spurious sprint aborts under sustained concurrent usage since some bd/Dolt call sites use `fatal: true` or lack `failSoft`.

Ready to merge.

PR #457, opened 2026-08-25
feat(memory-contract): extract memory-contract/v1 -- spec, schemas, bindings, conformance suite

ushah-blub0x:
## What this is

`memory-contract/v1` -- a written, machine-checkable description of the 23 memory and
code-intelligence tools (16 `kb_*`, 7 `code_*`): what they accept, what they return, and
every way they are allowed to refuse.

## Scope of behaviour change -- read this before reviewing

An earlier version of this description said "no runtime behaviour changes in this branch."
**That was wrong**, and a reviewer relying on it would mis-scope a 185-file diff. Corrected:

- **Inside `memory-contract/v1/`** -- no runtime behaviour change. It is all description,
  plus generated artifacts.
- **Riding along in the same branch** are four platform fixes and a trust payload that *do*
  change production behaviour. They are legitimate and tested, but they are not contract work:

| Path | Change |
|---|---|
| `src/services/http-transport.ts` | Blocked-port retry loop. Deterministic test in `tests/http-transport.test.ts`; fixed-port callers untouched. |
| `src/tools/compose-permissions.ts` | Exec timeouts 5s to 15s (onboarding path). |
| `packages/apra-fleet-se/apra-pm/` | Export-shrink guard in sprint dispatch, plus its `node:test` suite. |
| `scripts/run-all-tests.mjs` | Runner wiring for the above. |
| `.fleet/kb-canonical.json` | **+975/-163 -- 81 KB entries, all `CONFIRMED` (72 knowledge, 9 learning, 0 directives).** See below. |

Also touched: `.github/workflows/ci.yml`, `.gitignore`, `README.md`, `CHANGELOG.md`,
`llms-full.txt`, `package.json` / `package-lock.json`, `docs/`, and 12 unrelated test files.

### The bible update is a trust decision, not a data blob

`.fleet/kb-canonical.json` is the git-reviewed trusted import channel. **Merging this PR
blesses 81 `CONFIRMED` entries.** Verified composition: 72 `knowledge` + 9 `learning`, no
user-directives. Please review it as a deliberate step rather than scrolling past it.

## Four layers

| Layer | Where | Status |
|---|---|---|
| Prose spec | `memory-contract/v1/spec.md` | Sections 1-3 written (Envelope, Provider methods, Error model). Sections 4-5 are titled placeholders owned by T2/T3 by design. |
| JSON Schema 2020-12 | `memory-contract/v1/schemas/` | 46 files (request + response for each of 23 tools), generated. |
| Bindings | `memory-contract/v1/bindings/` | 23 MCP tool definitions + 1 OpenAPI RFC 9457 error catalogue. |
| Conformance hook | `memory-contract/v1/fixtures/`, `tests/` | 48 recorded fixtures, 11 test files / 75 tests, provider-parameterized round-trip harness. |

## Numbers

- 26 method entries: 12 `MemoryProvider` (P-1..P-12), 6 undeclared-but-tool-reachable
  (X-1..X-6), 1 property (X-7), 7 `CodeIntelligenceProvider` (C-1..C-7).
- 22 error codes in a **closed set** across 7 groups, plus 11 documented non-error outcomes
  and 3 deliberate exclusions.
- `npm run contract:check` on this HEAD: *23 tools, 46 schema files, 23 binding files,
  1 openapi file, 20 projectable taxonomy codes, all match and cross-reference cleanly.*
- `npx vitest run memory-contract`: **11 files, 75 tests, all passing.**

## Source of truth

Zod schemas in `src/tools/*.ts` are authoritative. `generate-contract.mjs` reads them from
`dist/tools/` and emits everything under `schemas/` and `bindings/`. The contract therefore
cannot describe a rule the code does not enforce. `contract:check` regenerates in memory and
compares byte-for-byte; CI runs it.

## Addressed from the code-level review of this PR

- **F-1 (high, contract defect) -- fixed.** `E-DIRECTIVE-QUARANTINE` was classified
  `surfaced: "response-field"` and projected to the wire with HTTP 403, but no response field
  reports it and the capture *succeeds*. An implementation built from that projection would
  refuse a directive capture this kernel accepts. Reclassified as a `non_error_outcome`
  alongside `E-CLAMP`, which it structurally matches. Closed set 23 to 22, projectable 21 to 20.
  Regenerated, not hand-patched -- bindings address codes by positional JSON pointer, so the
  governance refs in `kb_promote` and `kb_resolve_contradiction` renumber.
  A new falsifiability guard now requires any `response-field` code to name a `response_field`
  that exists in a raising tool's generated response schema. See `SIGNOFF.md` section 5.
- **F-2 -- fixed.** `spec.md` section 1 now states that the published schemas are deliberately
  wider (`minItems:1, maxItems:3`, `annotations`, `structuredContent`) than the reachable
  shape, and that consumers must validate against the schema, not the prose.
- **F-6 -- fixed.** `methods.json` `_meta` records why `kb_setup` has no provider-method entry.
- **F-3, F-5 -- addressed in this description** (the two sections above).
- **F-4 (`$id` pinned to `blob/main`) -- deferred to T11.** Both the `blob` and `raw` URLs 404
  today because the path is not yet on upstream `main`, and `blob/` serves HTML rather than
  JSON. Worth switching to a tag-pinned raw-content base before any external consumer exists.
  Not a merge blocker.

## Other things that look wrong and are not

1. **`spec.md` sections 4 and 5 are empty.** Reserved for T2/T3; the file states its own
   ownership rules. The nine invariant ids are already stamped as `x-invariant` annotations on
   28 generated schemas.
2. **`openapi.yaml` declares zero paths.** There is no HTTP route surface. It exists to publish
   the error vocabulary as RFC 9457 Problem Details. See `tests/BINDING-SCHEME.md`.
3. **The 7 `code_*` response schemas are unconstrained.** Deliberate -- those tools proxy to a
   swappable external provider whose payload shape we do not own. Request schemas *are* exact.
4. **`fixtures/` is excluded from the drift guard.** Recordings are not byte-reproducible
   (fresh UUIDs, wall-clock `created_at`, result order can flip). Kept honest by schema
   validation on every build instead, which is order- and id-agnostic. `README.md` item 5.
5. **Some `E-` names are not errors.** 11 outcomes deliberately carry no code (a confidence
   clamp is a *successful* capture with a reported adjustment). Each records why, and
   `tests/memory-contract-taxonomy.test.ts` enforces that a method never lists one among its
   `error_codes`.
6. **No code is retryable.** Transient failures are converted into actionable results at the
   provider boundary and never surface as errors.

## Pre-existing gaps this work surfaced (not introduced here)

- **6 methods are called by tools but absent from the declared `MemoryProvider` interface**
  (X-1..X-6). A second implementation built only from the interface would compile, then fail at
  runtime on `kb_list` / `kb_feedback`. Promoting them onto the interface is a separate change.
- **`repo_path` handling is not uniform.** `kb_export` / `kb_import` validate against the real
  filesystem and refuse pre-provider; `kb_session_prime` / `kb_stats` pass through verbatim.
  Recorded per tool as `validated-pre-provider` / `verbatim-passthrough`.
- **`HttpKbProvider` is dead code in this tree** (`INVENTORY.md` 4.4, F-10) -- worth knowing
  before any work that assumes an HTTP implementation already exists.

## Known limitations, stated on the record

- The drift guard's tool roster is hardcoded (`EXPECTED_TOOL_COUNT = 23`). It catches a tool
  *disappearing*; the newly-registered-but-unrostered case is covered by
  `tests/memory-contract-roster-guard.test.ts`.
- `tests/DEGRADATION.md` D-1..D-13 lists what schemas fundamentally cannot check
  (state-dependent refusals, cross-call ordering, filesystem side effects, provider identity,
  opaque `code_*` payloads). Handed forward to the behavioural test phase.

## Where to start reviewing

`memory-contract/v1/SIGNOFF.md` -- the per-layer verdict with cited evidence, plus section 5's
record of the F-1 correction and what it could *not* assert. Then `methods.json` and
`taxonomy.json`, the two hand-authored sources of truth. `schemas/` and `bindings/` are
generated; verify them with `npm run contract:check`, not by reading them.

## Conflicts with `main` are resolved

`main` was merged into this branch (not rebased -- 84 commits, and `pm` / `pm-kb` are
automation identities that a rebase would rewrite). Three files conflicted. Each resolution
and its reasoning:

### `.fleet/kb-canonical.json` -- took `main` wholesale, dropped this branch's export

This is the only file resolved by taking one side entirely, and the reasoning matters more
than the outcome.

Both sides are full snapshot exports, not appends: merge-base had 36 entries, this branch
exported 81, `main` reconciled down to 24. `main`'s reduction was deliberate and documented
in `docs/kb-bible-reconcile-2026-08-21.md`, which records the disposition of every drop.

Cross-checking this branch's 81 entries against that disposition: **15 entries existed at
merge-base, were dropped by the reconcile, and were still alive in this branch's export.**
14 are explained by name in the disposition doc:

| Disposition | Count | Ids |
|---|---|---|
| **Verified false on main** | 2 | `10e15056` (claims seven LLM providers including a `gemini` that does not exist), `735f21d3` (cites `resolveSecurePlaceholders`, a symbol absent from `src/`) |
| Contradiction loser | 1 | `ec8b983b` |
| Point-in-time sprint narration | 1 | `254b18e2` |
| Structural duplicate | 1 | `eb1611af` |
| Consolidated away | 4 | `1ae4ce00`, `66075242`, `b368ad28`, `daf6fb22` |
| Staled, never re-verified | 5 | `0633ac36`, `2ff14a5c`, `62a26c94`, `696d42bc`, `c996a17a` |

The 15th, `40295d4d`, asserts the embeddeddolt sandbox test can pass vacuously -- a hole the
disposition doc records as closed on main by `check-sandbox-sync-remote.test.ts:461-463`.

All 15 carried `confidence: CONFIRMED`. Merging them back would have re-blessed two claims
`main` verified false, on the git-reviewed trusted-import channel. This branch's export never
went through the verification gate that produced `main`'s 24; `main`'s did.

Nothing is lost. Those entries remain in the local KB, and the disposition doc gives the
correct recovery path: *"When that branch merges, run the ladder again; the entries need
re-capture, not just a sweep."* That re-capture is follow-up work, deliberately not folded
into this PR.

The resulting file is byte-identical to `main`'s, and its `provenance.entry_count` (24)
matches `entries.length` -- the self-consistency check whose absence the disposition doc
opens by describing.

### `CHANGELOG.md` -- kept both sides

Both sides append `## [Unreleased] -- <sprint>` blocks at the top, same convention. All three
of this branch's entries and all six of `main`'s are preserved. Ordered newest-first per the
file's convention: this branch's newest changelog commit is 2026-08-24 17:15, `main`'s is
2026-08-24 10:01.

### `tests/knowledge/kb-remote-member-e2e.test.ts` -- took the larger timeout

Both sides raised the same test's timeout for the same reason, from independent
measurements: this branch to 30000ms (3869ms observed in isolation, 77% of vitest's 5000ms
default), `main` to 20000ms (overruns under a full parallel run of 300+ files). Kept 30000ms
and both measurements in the comment.

### Post-merge verification

- Conflict-marker scan across every staged file, via PowerShell `Select-String` (not
  `findstr`, which fails silently on UTF-8): none.
- `npm run contract:check`: *23 tools, 46 schema files, 23 binding files, 1 openapi file,
  20 projectable taxonomy codes, all match and cross-reference cleanly.*
- The exact CI step -- `npm run contract:generate` then `git diff --exit-code` on `schemas/`
  and `bindings/` -- runs clean after the merge. Worth stating because the merge pulls in
  `chore: renormalize register-member.ts line endings to LF`, and this tree is generated on
  Windows while CI regenerates on `ubuntu-latest`.
- `git diff upstream/main --diff-filter=D`: **0 files** -- nothing from `main` was dropped.
- The merge touches **0 files** under `memory-contract/`.
- Nothing under `memory-contract/` references the Gemini provider that `main` removed.

## CI

This work was staged on a fork with no registered Actions workflows, so it carried zero
check-runs until now. This PR is the first occasion CI runs against it.

Local verification was done on Windows: the full vitest suite is **306 files passed / 8 skipped,
4173 tests passed**, with one pre-existing environmental failure --
`tests/undici-pack-install.test.ts` needs node-gyp to build `better-sqlite3`, and that host has
only the Microsoft Store python stub rather than a real interpreter. That test and its helper are
byte-identical before and after the merge, and `build-and-test (ubuntu-latest)` is green on
`main`, so it passes where Python exists.

---

https://claude.ai/code/session_013y1NpNfB8hRGasK2XKQeEY


kumaakh:
Thanks for the memory-contract extraction -- a few follow-ups:

1. **Non-ASCII in `tests/register-member-bootstrap-gate.test.ts`**: there's a literal checkmark emoji in a mocked return string (should be `[OK]` per repo convention). Note the installed pre-commit hook (`.github/hooks/pre-commit`) currently only enforces the ASCII check on `.yml/.yaml/.sh/.md` files -- `.ts`/`.js` are intentionally exempted (emoji in user-facing output is allowed there). So the hook wouldn't have caught this one on its own; worth running it anyway before pushing since it does catch the common cases, but this specific file falls outside its current scope -- please fix it by hand for now.

2. Please rebase onto latest `main` and resolve conflicts -- GitHub currently reports this PR as `CONFLICTING`. The PR description says conflicts were already resolved locally, but that resolution isn't reflected in a merge/rebase GitHub can validate, so a fresh sync is needed before this can merge.

3. As the memory-contract spec/schemas evolve, they need the same anti-drift discipline `apra-fleet-client` gets today (CLAUDE.md already calls that out for `src/tools/*` changes). Suggest adding an equivalent rule to CLAUDE.md: any future change to the memory subsystem (tools, schemas, bindings) must land together with the corresponding `memory-contract/v1` spec/schema/conformance updates in the same change, not as separate follow-up cleanup.

ushah-blub0x:
@kumaakh Thanks for the review -- all three addressed, head is now `90412ebd`.

1. **Non-ASCII checkmark** -- not fixed as asked: that glyph is the success sentinel `register-member.ts:536` matches via `startsWith`, so an ASCII marker makes the call read as "member not provisioned" and breaks 2 tests; I left it and added a comment explaining why, since a real fix means migrating the sentinel across `compose-permissions.ts` and the ~10 test files asserting on it.
2. **Conflicts** -- synced with latest `main` (167 commits) and resolved all 5 conflicts, including accepting your deletion of `docs/learnings.md`; GitHub now reports `mergeable: true`.
3. **CLAUDE.md anti-drift rule** -- added as one bullet right after the `packages/apra-fleet-client` one, requiring memory-subsystem changes (tools, schemas, bindings) to land with the matching `memory-contract/v1` spec/schema/conformance updates in the same change.

kumaakh:
Thanks Umang. I merged main into this branch and am realigning the contract with the current tool schemas.

One issue in the auto-sprint export guard (`auto-sprint.js:1317`, `lib/export-shrink-guard.mjs`) -- I'm fixing it on this branch:

- `git show HEAD:.beads/issues.jsonl` uses the default 1 MB `maxBuffer`. Past that it throws ENOBUFS, the empty `catch` reads it as "nothing committed", and the guard lets everything through. Our export is already 542 KB.
- A refusal exits 0 and prints only to the Haiku dispatch's stdout, which neither call site (`:3085`, `:3880`) reads. After one legitimate `bd delete`, every later sprint silently stops committing the export.

Fix: raise `maxBuffer` and fail closed on a read error; have the orchestrator detect a refusal and log it with the override, instead of leaving it in dispatch stdout. Pushing shortly.

PR #473, opened 2026-09-11
fix(vcs-auth): revoke the exact credential a deploy wrote, not an unlabeled/default-host guess

kumaakh:
## Summary

P0 fix for apra-fleet-7bjs, confirmed by two independent investigations (including an adversarial re-review).

- `scheduleCredentialCleanup`'s timer (`src/services/credential-cleanup.ts`) called `service.revoke(agent, cmds, exec)` with no `label`/`scopeUrl`. Since the credential FILE path is label-scoped but the git-config key is scopeUrl-scoped (`src/os/linux.ts`, `src/os/windows.ts`), this deleted the wrong (unlabeled, nonexistent) file while unsetting the RIGHT config key on the default deploy path -- git lost its credential registration while the real token file survived orphaned, unrevoked, on disk. A non-default `scope_url` made cleanup a silent no-op in the opposite direction.
- Fix: persist the label/scopeUrl actually used at deploy time on the agent record (`Agent.vcsCredentialLabel`/`vcsCredentialScopeUrl`, `src/types.ts`, set in `src/tools/provision-vcs-auth.ts`) and thread them back into `service.revoke()` from the cleanup timer, so cleanup always targets the exact credential-helper file/config-key pair the deploy wrote.
- Removed `credential-cleanup.ts`'s `DEFAULT_TTL_MS` (55-minute) fallback for an unknown expiry. `githubProvider.deployPat` returns no real `expiresAt`, so every PAT-mode deploy was scheduling its own near-arbitrary self-revoke ~55 minutes after deploy regardless of the token's real validity -- this explains "fresh" re-provisioned tokens failing minutes later. Chose to **skip scheduling entirely** when no real expiry is known, rather than probing GitHub's API for PAT expiry (a larger, riskier change). Tradeoff: a no-expiry credential now lives until explicitly revoked or re-provisioned with a real expiry, instead of auto-revoking at a made-up time. `checkVcsTokenExpiry`'s day-scale warning remains the reactive backstop.
- github-app mode is unaffected -- it always supplies a real `expiresAt`, so scheduling/threading behave as before, just correctly targeted now.

## Files changed
- `src/types.ts` -- add `vcsCredentialLabel`/`vcsCredentialScopeUrl` to `Agent`
- `src/tools/provision-vcs-auth.ts` -- persist label/scopeUrl at deploy time
- `src/services/credential-cleanup.ts` -- skip scheduling with no real expiry; thread label/scopeUrl into revoke
- `tests/credential-cleanup.test.ts` -- updated for new no-default-TTL semantics; added label/scopeUrl threading + multi-credential isolation regression tests
- `tests/provision-vcs-auth.test.ts` -- added label/scopeUrl persistence tests
- `tests/git-credential-helper-scoping.test.ts` (new) -- write/remove command-string parity + cross-label isolation for linux/windows OS command builders

`packages/apra-fleet-client` verified unaffected: no tool input schema changed (label/scope_url were already optional input fields; the fix is internal persistence/threading only).

## Test plan
- [x] `npm run build` -- clean
- [x] Targeted suites: `tests/credential-cleanup.test.ts`, `tests/provision-vcs-auth.test.ts`, `tests/revoke-vcs-auth.test.ts`, `tests/vcs-auth.test.ts`, `tests/register-member-vcs-provider.test.ts`, `tests/vcs-provider-detect.test.ts`, `tests/git-credential-helper-scoping.test.ts`, `tests/windows-credential-helper.test.ts` -- all pass
- [x] Full `npm test` -- 2803 pass, 0 fail

### DIFF
--- a/packages/apra-fleet-client/src/client/api.mjs
+++ b/packages/apra-fleet-client/src/client/api.mjs
@@ -227,7 +227,8 @@
  *   chosen when creating the token. Propagated to the member registry so provisioning can
  *   warn when the PAT is nearing expiry. Must be parseable by Date.parse -- the server
  *   REJECTS an unparseable value rather than storing it, because a NaN expiry silences the
- *   warning and makes the credential-cleanup timer fall back to its 55-minute default.
+ *   warning entirely (the credential-cleanup timer skips auto-revoke scheduling when no
+ *   real expiry is known, rather than falling back to any default TTL).
  */
 
 /**
--- a/src/services/credential-cleanup.ts
+++ b/src/services/credential-cleanup.ts
@@ -7,8 +7,6 @@ import { bitbucketProvider } from './vcs/bitbucket.js';
 import { azureDevOpsProvider } from './vcs/azure-devops.js';
 import type { VcsProviderService } from './vcs/types.js';
 
-const DEFAULT_TTL_MS = 55 * 60 * 1000; // 55 minutes
-
 // apra-fleet-5co8.5.1: setTimeout's delay is a signed 32-bit int internally;
 // Node (and browsers) SILENTLY CLAMP an overflowing delay to ~1ms rather than
 // running it after the full requested duration (see Node's lib/timers.js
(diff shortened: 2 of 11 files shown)


kumaakh:
## Adversarial review follow-up

Ran the fix through an independent Opus adversarial review (the same reviewer that originally confirmed/scoped this P0). Verdict: the reported timer-path bug was genuinely fixed (persistence, arg-order across all 3 providers, and the skip-scheduling logic all checked out), but it found two additional occurrences of the same bug class that the original fix missed:

1. **`src/tools/remove-member.ts:80`** — member decommissioning called `vcsService.revoke(agent, cmds, exec)` with no label/scopeUrl, same unlabeled/default-host guess as the original bug, but here on the machine actually being removed from the fleet (leaving a live, unrevoked token file behind). Fixed to pass `agent.vcsCredentialLabel`/`agent.vcsCredentialScopeUrl`.
2. **Superseded-credential orphan**: deploying a new credential under a *different* label than a previously provisioned one silently orphaned the old one -- `cancelCredentialCleanup()` discards the prior timer (the agent record only tracks one active label/scopeUrl pair), so the superseded credential was never revoked and never rescheduled. `provision-vcs-auth.ts` now explicitly revokes a superseded credential (different label or scopeUrl) before deploying the new one; a same-label refresh is unaffected since the write already overwrites in place.

Also fixed: a misleading comment claiming `checkVcsTokenExpiry` provides a reactive backstop for the no-expiry/skip-scheduling case (it doesn't -- it's only invoked when an `expiresAt` is already known), and several stale "falls back to a 55-minute default" comments left over from the removed `DEFAULT_TTL_MS`.

New regression tests: `remove_member` passes persisted label/scopeUrl through to revoke; a label-changing re-provision revokes the superseded credential; a same-label re-provision does not self-revoke. Also retitled/clarified the OS-command-builder test file to be explicit that it's characterization coverage (it would pass pre-fix too), not the actual caller-side regression coverage, and replaced a tautological assertion in `credential-cleanup.test.ts` with a real two-member isolation scenario.

**Test results after these follow-ups:**
- `npm run build` -- clean
- Targeted suites (9 files, 187+27 tests across two runs) -- all pass
- Full `npm test` (vitest root + apra-fleet-se workspace): apra-fleet-se 2803/2803 pass; root vitest 4514 pass / 1 failed / 41 skipped -- the single failure (`tests/undici-node20-compat.test.ts`) plus one failed-suite (`tests/gen-sea-config.test.ts`, 4 skipped) are both pre-existing environment issues in this worktree (`node_modules/ajv` missing), confirmed unrelated to this diff by inspecting the actual error output.

PR #490, opened 2026-09-17
fix(fleet-se): fix flaky serve-wiring-integration backlog render assertion

kumaakh:
The GET / integration test in serve-wiring-integration.test.mjs polled only for a 200 response before asserting on rendered dashboard markers. A 200 does not guarantee renderBeadsHtml's async bd read has finished, so the first successful response could still be missing `id="backlog-table"`, failing the marker assertion even though the server was healthy.

Confirmed as a pre-existing flake on main, not caused by any specific PR's diff: it hit CI independently on two unrelated branches (PR #488, docs-only playbook addition; PR #489, this same runbook doc) -- neither touches serve.mjs/dashboard rendering. Fixed by polling on the actual rendered marker instead of just the status code. Same fix already applied directly to both of those PR branches; landing it on main too so future branches don't rediscover it.

Verified with a narrow run of just this file: `node --test test/serve-wiring-integration.test.mjs` -- 4/4 passing.


kumaakh:
Superseded: the marker-poll fix this PR carried is already merged into main via #491 (which also includes the actual root-cause fix, apra-fleet-4ipl). No remaining diff to land.

PR #497, opened 2026-09-21
feat(fleet-bridge): DevOps-to-fleet-sprint bridge, proven on a live Azure DevOps sprint

kumaakh:
## Status: proven end to end against a live Azure DevOps project

Every requirement in the original brief now works against `apralabs/e2e-fleet-testing`, verified by running it rather than by unit tests:

| Requirement | Evidence |
|---|---|
| Runner prepared for fleet-sprints | `preflight`, nine checks |
| Trigger selecting work items, target branch, goal | `ingest` -> `launch` |
| Sprint develops, reviews, tests | Plan, Develop, Review, Deploy, Integ Test, Final Review, Harvest |
| **Raises a PR against the target branch** | PR !3 in the toy repo, 10 commits, +547/-6 |
| **Carry-over becomes new backlog items** | Work items #5-#8 and #9-#12, created by the bridge |
| Beads closed during the sprint are NOT reported | Verified: the closed originals were excluded |
| Progress observable from logs | `status` reports phase, health, verdict, spend, verbatim terminal reason |
| Long sprints do not hold runner slots | Detached dispatch; `launch` returns in seconds |

Carry-over publication is idempotent: a re-run creates nothing and exits 0, so a pipeline retry is safe.

## Defects found by running it -- none of which the unit suite could see

Seventeen in total across three live sprints. The suite was green throughout. Every one was a mismatch with a real external surface, and the pattern is recorded in `packages/apra-fleet-se/docs/fleet-bridge-build-log.md`: **a fake built from the calling code's assumptions tests the assumption, not the system.**

The ones worth a reviewer's attention:

- **Command injection.** `buildTrackerCommand` interpolated an externally-supplied credential NAME bare into a shell command dispatched to a member, gated only on truthiness. A name containing `}` closed the placeholder and the remainder executed on the member host, reachable by anyone able to queue a pipeline. `rest-client.mjs` gated the identical interpolation using its own duplicate copy of the charset -- which is how the two drifted. One validator now, both call sites.
- **The bridge and the engine were never connected.** `--synthetic-root-id`, the beads root that `ingest` exists to produce, was parsed, validated, stored in the handle, and never placed in the launch body.
- **The PAT secret name was unplumbable**, so the engine fell back to a default secret belonging to an unrelated project -- failing the PR *and* overwriting the member's working git credential. Fixed in four call sites; a source-scan guard now fails if any call site omits it.
- **Carry-over could never work.** `bd ado push` sets `System.State='New'`, an Agile/Scrum state a Basic-process project rejects with a 400 -- reported as a Warning with exit 0, so three layers read it as success. Creation now goes over REST setting no state at all, so the tracker applies each type's own initial state.
- **A guard test that could not fail.** The error-rule guard passed while a live violation sat in the tree. Rewritten and falsified against each violation class.
- **`spool.claim()`'s single-writer invariant was in-process only** -- two daemons starting together both won, the exact case its comment says corrupts the append-blob cursor.

## Known gaps, deliberately not fixed

- The append-blob sink, archive exporter and live SPA are built and tested but wired to nothing.
- `fleet-bridge viewer` binds `0.0.0.0` with no client auth and injects the supervisor bearer. **Accepted risk**, documented in `docs/setup.md` with the reasoning; a cookie-based guard is the intended design.
- The PowerShell branch of the REST transport is unproven -- the test member is gitbash, so only the POSIX path has executed.
- GitHub and Bitbucket adapters are not written. The abstraction exists; only Azure DevOps is proven.

## Tests

1035 bridge tests passing. `apra-fleet-se` at 3294 passing with 3 pre-existing Windows-only failures unrelated to this branch. Secret scan clean: zero gitleaks findings inside this diff.

---

## Update: review addressed (commit 2 of 2)

All eight findings triaged. **Each was verified against the code before being acted on, which changed five of them.** The review was specific and mostly right; the corrections below are offered as findings in their own right, not as pushback.

| # | Review said | Verified | Action |
|---|---|---|---|
| 1 | Gate is "backwards"; ad hoc watch ungated | **Over-broad, not inverted.** Ad hoc watch IS correctly gated and must stay so. The daemon reads its OWN claim as foreign and drops the append-blob sink on every sprint | Fixed -- identity compared, fails safe to "foreign" |
| 2 | Missing error listener crashes the daemon | Confirmed exactly | Fixed at stream creation |
| 3 | Comment failures are silent | Confirmed end to end -- no log line anywhere | Fixed in the facade |
| 4 | 15-min delay then WATCH_LOST | **Worse.** `gotData` resets each tick, so it polls FOREVER and the daemon never finalizes | Fixed via a shared classifier |
| 5 | Operator gets a 404 | **Latent only** -- `finalize.mjs:575` guards on `archive.ok` | Fixed anyway; return value was self-inconsistent |
| 6 | Silent POSIX default | True, but the adjacent member-less path returns nulls **deliberately** | Only the catch fixed |
| 7 | Branding; client needs updating | Branding true. **Client concern unfounded** -- supervisor HTTP field, not an MCP tool schema | Renamed; old spelling accepted one release |
| 8 | `allSettled` fixes latency | Independent, yes -- but MAX-dominated by one `bd` subprocess | **Declined on evidence** |

### Four defects the review missed

The most serious: **`jsonl-file` published no `health()`**, and the monitor only watches sinks that do. Combined with finding 1, a daemon could be **writing nothing anywhere while reporting healthy**. Also: `stop()` never awaited `end()` (lost buffered lines on fast shutdown); the sink gate is evaluated only at startup; and `setBuildStatus()` POSTs to the create-build endpoint (left alone -- no verb calls it).

### Beyond the review: repeat-run behaviour

Tracing what happens when a pipeline fails and is re-run found that **the only machine-independent control in the entire flow is `external_ref` on a bead**, which the bridge consumes as a beads guarantee without verifying. Two fixes:

- `ingest` refuses to reparent a bead whose current root is **reserved by a live sprint**. The supervisor re-expands scope by live BFS from its issue roots, so a re-ingest with a changed ref list silently shrank a running sprint -- and the overlap guard could not see it, because it is keyed on roots and reparenting is the move that relocates beads between them. An unreachable supervisor now stops the run rather than collapsing into "no reservations exist".
- A closed root is no longer invisible to the reuse lookup (`bd list` defaults to open only). Only-closed matches **refuse** rather than revive or duplicate.

Still true and documented, not fixed: `launch` has no dedupe keyed on work items, and a crash between carry-over's push and its `external_ref` stamp can duplicate on retry.

### The boundary checker mattered more than the rename

`--vcs-pat-secret-name` / `vcs_pat_secret_name` drop the brand; the supervisor accepts the old spelling for one release with a deprecation warning, so nothing migrates in lockstep. The ~60 internal sites keep their name -- not a public surface.

The underlying problem was that `check-generic-boundary.mjs` **could not have caught this**: its file set excluded `bin/` and `src/supervisor/` entirely, and it had no provider-branding rule. Both widened. Of 35 new hits, 5 were reworded to remove the assumption and 30 are anchored, dated, per-pattern exceptions with reasons. No file-level exemptions.

### Tests

1125 bridge (was 1078), 0 failing. apra-fleet-se 3297 passing with the 3 known Windows-only failures. **Every fix carries a test proven to fail against the previous code** -- verified by restoring the pre-fix file and re-running, since several of these findings exist precisely because a passing test proved nothing.


packages/apra-fleet-bridge/src/verbs/watch.mjs
```diff
@@ -0,0 +1,558 @@
...
+  if (!deps.spool) {
+    throw new BridgeError(
+      BRIDGE_ERROR_CODES.CONFIG_MISSING,
+      'watch: deps.sinks includes a shared:true sink but no deps.spool was provided to check for a live claim',
+      { param: 'deps.spool' }
+    );
+  }
+
+  const doc = await deps.spool.read(opts.sprintId);
+  const liveClaimExists = Boolean(doc && doc.claim);
```
kumaakh: **Risk:** `applySinkGate()` disables every `shared:true` sink whenever ANY claim exists, without checking if the claim is the caller's own. `daemon.mjs` always self-claims before calling `runWatch`, so this strips the durable sink from the legitimate owner while an unclaimed ad hoc run sails through unprotected -- inverted protection.

**Suggested change:** Compare the claim's `{pid, host}` against the caller's own identity; only gate the sink when the claim belongs to a different pid/host.

**Options considered:** (a) drop-then-reclaim around `runWatch` -- reopens the race the claim prevents; (b) thread an `isSelf` flag from `daemon.mjs` -- duplicates identity logic per call site; (c) compare identity inside `applySinkGate` using the same shape `spool.claim()` already writes.

**Why (c):** fixes the bug where the identity data already lives, with no new call-site plumbing, and is correct in both directions (self vs. foreign claim).

packages/apra-fleet-bridge/bin/runtime.mjs
```diff
@@ -0,0 +1,210 @@
...
+ * it only ever calls `.write(chunk)` (in `emit()`) and `.end()` (in
+ * `stop()`, guarded by `typeof stream.end === 'function'`) on whatever this
+ * returns -- both of which a real `fs.WriteStream` provides natively (it is
+ * also a full `EventEmitter`, so `.on(...)` is available if a future caller
+ * needs it, but nothing in this package currently does).
+ *
+ * @param {string} path
+ * @returns {import('node:fs').WriteStream}
+ */
+export function openAppendStream(path) {
```
kumaakh: **Risk:** No `'error'` listener anywhere on this `fs.createWriteStream`. A disk-full (ENOSPC) or permission error during a long run throws Node's default uncaught-error path, crashing the whole daemon -- including healthy sinks -- contradicting this package's own per-sink-isolation contract (`sinks/health.mjs`).

**Suggested change:** Attach an `'error'` listener at stream-creation time that marks this sink degraded via the existing health-reporting mechanism instead of letting the error propagate.

**Options considered:** (a) try/catch around `write()` calls -- doesn't help, `'error'` is emitted async, not thrown; (b) listener here that routes into the existing degraded-sink path; (c) (b) plus one reopen attempt before giving up.

**Why (b):** only option that actually catches the async error, and reuses the isolation mechanism already documented for every other sink.

packages/apra-fleet-bridge/src/adapters/azure-devops.mjs
```diff
@@ -0,0 +1,655 @@
...
+  if (workItemId === undefined || workItemId === null || workItemId === '') {
+    throw new BridgeError(
+      BRIDGE_ERROR_CODES.CONFIG_INVALID,
+      `azure-devops adapter: comment() requires opts.workItemId, got ${JSON.stringify(workItemId)}`,
+      { field: 'workItemId', value: workItemId }
+    );
+  }
+  const secretName = r.adoPatSecretName || DEFAULT_ADO_PAT_SECRET_NAME;
+  const url = `${trimTrailingSlash(r.adoOrgUrl)}/${encodeURIComponent(r.adoProject)}/_apis/wit/workItems/${encodeURIComponent(String(workItemId))}/comments?api-version=7.1-preview.3`;
+  return deps.restClient({
```
kumaakh, line 594: **Risk:** `comment()` (and `setBuildStatus()` at line 631) return `deps.restClient(...)`'s result directly with no status-check, unlike `createWorkItem()` in this same file. `facade.mjs` wraps `adapter.comment()` in try/catch specifically to log-and-swallow failures, but since this never throws on 404/401/403, that catch never runs -- comment failures are completely silent.

**Suggested change:** Route both through the same `parseRestJson`-style status check `createWorkItem()` already uses, so a non-2xx throws and the caller's existing catch-and-log actually fires.

**Options considered:** (a) inline ad hoc status checks -- duplicates existing centralized logic; (b) reuse `parseRestJson`, matching this file's own convention; (c) push status-checking into `facade.mjs` instead -- leaks provider-specific response shape into the generic facade.

**Why (b):** one-line convention match against code already in this file; keeps provider-specific handling in the provider file.

kumaakh (commented):
## Risk review: PR #497 vs existing fleet-sprint

Since apra-fleet-bridge is the caller that *creates* fleet-sprint runs (not the reverse), reliability bugs in the bridge's watch/comment/archive paths directly affect the operational experience of every sprint launched through it, not just new opt-in code. Full findings below, each as: **Risk -> Suggested change -> Options considered -> Why this approach**. Findings 1-6 are also left as inline comments on their exact lines.

---

### 1. Sink-gate inversion defeats single-writer protection in the primary daemon path (`watch.mjs:260`)
**Risk:** `applySinkGate()` disables every `shared:true` sink whenever ANY claim exists on the spool doc, without checking whether the claim belongs to the calling process. `daemon.mjs` always self-claims before calling `runWatch`, so the gate reads its own claim as "someone else owns this" and strips the durable append-blob sink from the intended owner -- while an unclaimed ad hoc `fleet-bridge watch` sails through unprotected. The protection is backwards.

**Suggested change:** Compare the claim's `{pid, host}` against the caller's own identity before treating it as a foreign claim; only gate the sink when the claim belongs to a *different* pid/host.

**Options considered:** (a) Drop the self-claim before calling `runWatch` and re-claim after -- rejected, reopens the race the claim exists to prevent. (b) Pass an `isSelf` flag down from `daemon.mjs` -- works but duplicates identity logic at each call site. (c) Compare claim identity inside `applySinkGate` itself using the same `{pid, host}` shape already written by `spool.claim()` -- single source of truth, no new call-site plumbing.

**Why (c):** It fixes the bug where the identity data already lives (the claim record itself) rather than pushing the comparison outward to every caller, and it makes the ad hoc-vs-daemon distinction correct in both directions with one change.

---

### 2. No `error` listener on the append-stream: a disk-full event can crash the whole daemon (`runtime.mjs:208`)
**Risk:** `openAppendStream()` returns a raw `fs.createWriteStream` with no `'error'` listener anywhere in the chain. An ENOSPC or permission/mount error during a multi-day run throws Node's default uncaught-error path and kills the entire fleet-bridge daemon, taking down every sink -- including healthy ones -- contradicting this package's own documented per-sink-isolation goal (`sinks/health.mjs`: "a sink must never kill the process").

**Suggested change:** Attach an `'error'` listener where the stream is created that routes into the same per-sink health/degradation reporting the other sinks already use, instead of letting it propagate.

**Options considered:** (a) Wrap every `stream.write()` call site in try/catch -- doesn't help, `'error'` on a WriteStream is emitted asynchronously, not thrown synchronously from `write()`. (b) Attach the listener in `jsonl-file.mjs` at stream-creation time and mark that sink degraded on error, matching the pattern other sinks use. (c) Same as (b) but also close and attempt to reopen the stream once before giving up.

**Why (b) [with (c) as a nice-to-have]:** It's the only option that actually catches the async error event, and it reuses the existing degraded-sink mechanism rather than inventing a new failure path -- consistent with the file's own stated isolation contract.

---

### 3. Azure DevOps adapter's `comment()`/`setBuildStatus()` swallow failures the caller is explicitly trying to catch and log (`azure-devops.mjs:594`, `:631`)
**Risk:** Both functions return `deps.restClient(...)`'s result directly without status-checking, unlike `createWorkItem()`/`resolveWorkItemType()` in the same file which do. `facade.mjs` wraps `adapter.comment()` in try/catch specifically to log-and-swallow ("COMMENT NEVER THROWS"), but since a 404/401/403 never throws here, that catch never fires -- not even the promised non-fatal log line appears. Every progress-comment failure during a sprint is completely silent.

**Suggested change:** Route both functions through the same `parseRestJson`-style status check `createWorkItem()` already uses, so a non-2xx response throws and the caller's existing catch-and-log actually runs.

**Options considered:** (a) Add ad hoc status checks inline in each function -- works but duplicates logic already centralized in `parseRestJson`. (b) Reuse `parseRestJson` for both, matching the file's own existing convention. (c) Leave as-is and add status checking in `facade.mjs` instead -- rejected, pushes provider-specific REST-response shape knowledge into the generic facade.

**Why (b):** It's a one-line convention match against code already in the same file, and it keeps provider-specific response handling inside the provider file rather than leaking it upward.

---

### 4. `getLog()` fallback never classifies terminal failures, unlike the already-fixed equivalent in `status.mjs` (`watch.mjs:462`)
**Risk:** When `getSprint` returns null (sprint dropped from the supervisor's ledger before going terminal) but `getLog` still succeeds with a log containing the engine's own "Sprint failed: ..." marker, this fallback always reports `health:'unknown'` instead of calling `classifyTerminalLog()`/`buildTerminalFromEvidence()`. The loop's only exit condition is unreachable on this path, so watch polls for the full `giveUpMs` (15 min default) before throwing a generic `WATCH_LOST`, discarding the real failure reason and delaying `daemon.mjs`'s finalize handling by up to 15 minutes.

**Suggested change:** Mirror `status.mjs:263-266`'s fix -- call the same terminal-log classification in this fallback path instead of hardcoding `health:'unknown'`.

**Options considered:** (a) Extract the classification call into a shared helper used by both `status.mjs` and `watch.mjs` -- best long-term, avoids drift the next time this logic changes. (b) Copy the fixed snippet verbatim into `watch.mjs` -- fixes the immediate bug with minimal diff. (c) Leave `watch.mjs` as-is and shorten `giveUpMs` -- rejected, treats the symptom (long wait) not the cause (wrong classification), and still discards the real failure reason.

**Why (a), or (b) if scope is tight for this PR:** This exact incident was already fixed once in `status.mjs`; letting `watch.mjs` regress the same bug independently is the actual root risk. A shared helper prevents a third occurrence; a verbatim copy at minimum stops this one from currently existing.

---

### 5. `archive-publisher.mjs` can report a valid `indexUrl` when `index.html` itself failed to upload (`archive-publisher.mjs:146`)
**Risk:** `indexUrl` is set whenever `uploaded > 0`, not specifically when `index.html` uploaded. If `index.html`'s upload fails (transient 5xx) while a later file succeeds, `uploaded` is still >0, so `indexUrl` is set and returned as valid even though `ok` is false -- an operator following the link gets a 404 despite the result object claiming success.

**Suggested change:** Only populate `indexUrl` when the specific `index.html` upload's own result is successful, independent of the aggregate `uploaded` counter.

**Options considered:** (a) Check the aggregate `ok` flag before setting `indexUrl` -- doesn't fully fix it, since `ok` could be computed from a different failure. (b) Track the `index.html` upload's own success/failure explicitly and gate `indexUrl` on that specific result. (c) Always compute `indexUrl` unconditionally and let the caller check `ok` -- rejected, this is exactly the current bug's shape (a value present alongside a false `ok`), just moved, not fixed.

**Why (b):** It ties the URL's validity to the one upload result it actually depends on, rather than an aggregate counter that can be true for unrelated reasons.

---

### 6. `memberDialectFor()` silently defaults to POSIX on lookup failure -- violates this repo's own hard-fail rule (`fleet-bridge.mjs:305`)
**Risk:** On any `memberDetail` lookup failure, this returns `{targetOs: null, shell: null}` and only logs a warning; `getSeCommands()` then treats a null OS as POSIX. Per this repo's own CLAUDE.md rule ("A POSIX-only feature must hard-fail on Windows or gate with a surfaced error -- an advisory warning that never blocks is a false success"), a transient lookup failure for a PowerShell member causes the bridge to silently dispatch POSIX-shell commands against it, producing opaque failures instead of a clear, blocking error.

**Suggested change:** On lookup failure, surface a blocking error (or a typed "unknown dialect, cannot proceed" result the caller must handle) instead of defaulting silently to POSIX.

**Options considered:** (a) Retry the lookup once before failing -- helps transient cases but still needs a hard failure path for persistent ones, so doesn't replace the fix, only supplements it. (b) Default to whichever OS the bridge itself runs on -- rejected, no better justified than defaulting to POSIX and just moves the wrong guess. (c) Fail closed: surface a blocking error naming the member and the lookup failure, exactly matching this repo's existing rule for POSIX-only features.

**Why (c):** It's precedent already established in this exact codebase for exactly this failure shape (advisory-warning-as-false-success is explicitly called out as not acceptable), so this is a straightforward consistency fix, not a new judgment call.

---

### 7. Generic engine boundary violation: Azure-DevOps-branded secret-name parameter, pre-existing on `main` and extended by this PR (`bin/cli.mjs:181` and 9 files total)
**Risk:** `azdevopsPatSecretName` already existed inside `vcs-auth.mjs` on `main` before this PR (its own header already promises "this function -- and runner.js as a whole -- never names or branches on a provider", so the debt predates this PR). PR 497 extends and deepens it: it adds the new public `--azdevops-pat-secret-name` CLI flag (`bin/cli.mjs:181`) and the new `azdevops_pat_secret_name` supervisor-API JSON field (`src/supervisor/api.mjs`), and threads it into new call sites in `abort.mjs`, `publish-pr.mjs`, and `spawner.mjs` -- all part of wiring the new bridge-to-sprint launch path. The need is genuine (a hardcoded secret name is unworkable when multiple stored credentials could satisfy a workflow's need), but the naming bakes today's one PAT-based provider into the generic engine's public surface, and this PR is what turns it from an internal parameter into a public, documented (`--help` text says "...holding the Azure DevOps...") CLI/API contract. The day a second PAT-based provider needs the same override, that force either a misnomer or a breaking rename across all 9 files -- and a breaking rename is much more costly once it's a public flag than while still internal.

**Already-generic internally, so this is purely a naming/interface fix, not a redesign:** tracing the parameter through `vcs-auth.mjs` shows `provisionVcsAuthForMember` already forwards it as an opaque generic `secretName` into `buildProvisionArgsForProvider({..., secretName})`, which hands it to `impl.buildProvisionArgs({base, repoRef, availableSecrets, secretName})` -- a per-provider hook. Only `vcs-providers/azure-devops.mjs` actually reads `ctx.secretName` (falling back to its own `DEFAULT_PAT_SECRET` otherwise); every other provider already ignores it harmlessly. The provider-specific branding exists ONLY at the public boundary (CLI flag text, JSON field names, function parameter names) -- the actual plumbing is provider-agnostic already.

**Suggested change -- generalize the public boundary to a named credential-need mechanism, in two parts:**
1. Rename the single-value threading end-to-end: `azdevopsPatSecretName` -> `vcsPatSecretName`, `--azdevops-pat-secret-name` -> `--vcs-pat-secret-name`, `azdevops_pat_secret_name` -> `vcs_pat_secret_name`. Zero internal logic change -- `secretName` inside `vcs-auth.mjs` is unchanged, only the public-facing names lose the provider brand.
2. For the deeper, stated goal ("inform the workflow which of the existing secrets represents its need(s), generically"): let each VCS provider module declare what credential need(s) it has via a new optional hook, e.g. `credentialNeeds()` returning `['pat']` (azure-devops) or `[]` (github, which uses github-app auth with no named-secret need). The engine-facing parameter becomes a small map keyed by need, e.g. `credentialSecretNames: { pat: 'fleet_bridge_azdevops_pat' }`, populated via a repeatable CLI flag `--credential-secret <need>=<name>` (e.g. `--credential-secret pat=fleet_bridge_azdevops_pat`). `buildProvisionArgsForProvider` forwards the whole map (or just the value the provider's own declared need(s) select) to `impl.buildProvisionArgs`, exactly as it forwards `secretName` today -- no new branching in the generic files.

**Options considered:**
- **(a) Rename only (part 1), stop there.** Smallest possible diff, zero risk, fixes today's literal violation. Leaves the underlying "one string, one hardcoded meaning" shape in place -- the next PAT-based provider would need a second generic-sounding-but-still-singular field (e.g. `vcsPatSecretName` doesn't generalize to a provider needing two different named secrets, like a PAT plus a separate webhook signing secret).
- **(b) Full generic credential-needs map (parts 1+2).** Matches the pattern this codebase has already established elsewhere for exactly this class of problem: `docs/adr-taskdb-backend-neutral-interface.md`'s VCSModule work, and the existing `apra-fleet-647.1.5` line of fixes that pulled a provider-keyed `DEFAULT_AUTH_MODES` table out of `vcs-module.mjs` for the identical reason (a provider-keyed structure living in supposedly generic code). This PR's `azdevopsPatSecretName` is the same anti-pattern recurring in new code.
- **(c) Move secret selection out of fleet-sprint entirely, into the credential store itself** (e.g. a `role` or `tag` on each stored secret that a provider looks up by role rather than by an operator-supplied name at all). Most "automatic" long-term, but changes the credential-store's own data model and is a much larger, cross-cutting change than this PR's scope justifies right now.

**Why (b):** It solves the actual generic need you're describing (a workflow needs to say which stored secret satisfies which of its credential requirements, without any engine file ever hardcoding a provider name) using a mechanism this codebase has already validated for the same problem shape elsewhere, at a cost of touching the same 9 files this PR already touches -- no larger a diff than what's already here, just correctly shaped. (a) is an acceptable smaller stopgap if timeline pressure requires it, with (b) filed as an immediate follow-up rather than left as unaddressed debt; (c) is worth a separate design pass later but shouldn't block this PR.

---

### 8. `runPreflight` runs 6 independent checks sequentially despite the file's own stated goal of running all checks without short-circuiting (`preflight.mjs:523`)
**Risk:** Member fetch, VCS provider check, credential check, repo/base check, beads-health dolt-pull probe, and playbook check are all `await`-ed one after another even though they touch disjoint backends (fleet API, filesystem/git, beads) with no shared state. Total latency is the SUM of each check's I/O time instead of the MAX, inflating wait time on every single preflight run.

**Suggested change:** Run the independent checks concurrently via `Promise.allSettled`, then assemble the existing pass/fail report from the settled results in the same order/shape as today.

**Options considered:** (a) `Promise.all` -- rejected, a single check throwing would abort the others mid-flight and lose their results, defeating the file's own "never short-circuit" goal. (b) `Promise.allSettled` over all 6 checks. (c) Partial parallelism (e.g. group only the checks with zero interdependency) -- unnecessary complexity here since the review found no shared state between any of the 6.

**Why (b):** It's the direct concurrency-safe equivalent of "run every check, don't stop early," which is already this file's documented intent -- `allSettled` guarantees every check completes and reports regardless of any other check's outcome, cutting wall-clock time to roughly the slowest single check.

packages/apra-fleet-bridge/src/verbs/watch.mjs
```diff
@@ -0,0 +1,558 @@
...
+        } catch (err) {
+          log(`[watch] getLog("${sprintId}") fallback failed: ${safeMessage(err)}`);
+          logCallFailed = true;
+        }
+        if (!logCallFailed) {
+          // The supervisor answered (even a 404-as-null "no log file yet" is
+          // an answer, per supervisor-client.mjs) -- that is reachability,
+          // whether or not there is any text to show.
+          gotData = true;
+          snapshot = buildFallbackSnapshot(sprintId, logTail ?? null, now());
```
kumaakh: **Risk:** This `getLog()` fallback (used when `getSprint` returns null) always reports `health:'unknown'` and never calls `classifyTerminalLog()`/`buildTerminalFromEvidence()`, unlike `status.mjs`'s already-fixed fallback for the identical incident. The loop's only exit condition becomes unreachable on this path, so watch polls the full `giveUpMs` (15min default) before a generic `WATCH_LOST`, discarding the real failure reason and delaying `daemon.mjs`'s finalize by up to 15 minutes.

**Suggested change:** Mirror `status.mjs`'s fix -- call the same terminal-log classification here instead of hardcoding `health:'unknown'`.

**Options considered:** (a) extract a shared helper used by both files -- best long-term, prevents future drift; (b) copy the fixed snippet verbatim -- minimal diff, fixes this instance; (c) just shorten `giveUpMs` -- treats the symptom, still discards the real failure reason.

**Why (a), or (b) if scope is tight:** this exact bug was already fixed once elsewhere; a shared helper stops it from recurring a third time.

packages/apra-fleet-bridge/src/spa/archive-publisher.mjs
```diff
@@ -0,0 +1,166 @@
...
+          }
+          reason = `status ${status}`;
+        } catch (err) {
+          reason = safeMessage(err);
+        }
+        failures.push({ path: file.path, reason });
+      }
+
+      // The unsigned URL, on purpose -- see this module's header.
+      const indexUrl = uploaded > 0 ? `${base}/${containerName}/${prefix}/index.html` : null;
```
kumaakh: **Risk:** `indexUrl` is set whenever `uploaded > 0`, not specifically when `index.html` uploaded. If `index.html` fails (transient 5xx) while a later file succeeds, `indexUrl` is still set and returned as valid despite `ok` being false -- a caller following the link gets a 404 despite a non-null `indexUrl`.

**Suggested change:** Gate `indexUrl` on the `index.html` upload's own specific result, not the aggregate `uploaded` counter.

**Options considered:** (a) gate on aggregate `ok` instead -- doesn't fully fix it, `ok` could be false for an unrelated file; (b) track `index.html`'s own success explicitly; (c) always set `indexUrl` and let the caller check `ok` -- this is the current bug's shape, just relocated.

**Why (b):** ties the URL's validity to the one upload result it actually depends on.

packages/apra-fleet-bridge/bin/fleet-bridge.mjs
```diff
@@ -0,0 +1,1207 @@
...
+    if (!memberName) return { targetOs: null, shell: null };
+    try {
+      const { fleetApi } = await mcp();
+      const raw = await fleetApi.memberDetail({ member_name: memberName, format: 'json' });
+      const detail = parseToolJson(raw);
+      const targetOs = (detail && (detail.os || detail.platform)) || null;
+      const shell = (detail && detail.shell) || null;
+      return { targetOs, shell };
+    } catch (err) {
+      log(`[fleet-bridge] could not resolve member "${memberName}"'s os/shell (defaulting to POSIX dialect): ${safeMessage(err)}`);
```
kumaakh: **Risk:** On any `memberDetail` lookup failure this silently returns `{targetOs: null, shell: null}` with only a warning log; `getSeCommands()` treats null OS as POSIX. Per this repo's own CLAUDE.md rule ("a POSIX-only feature must hard-fail on Windows... an advisory warning that never blocks is a false success"), a transient lookup failure for a PowerShell member causes silent POSIX-shell dispatch against it.

**Suggested change:** Surface a blocking error (or a typed unknown-dialect result the caller must handle) on lookup failure instead of defaulting to POSIX.

**Options considered:** (a) retry once before failing -- helps transient cases but still needs a hard-fail path for persistent ones; (b) default to the bridge's own host OS -- equally unjustified, just a different wrong guess; (c) fail closed with a named error, matching this repo's existing rule for exactly this failure shape.

**Why (c):** precedent already established in this codebase for this exact failure shape -- a consistency fix, not a new judgment call.

packages/apra-fleet-se/bin/cli.mjs
```diff
@@ -168,6 +168,17 @@ export function buildOptionsSpec() {
...
+        // 'azdevops_pat') is just one convention, not a guarantee -- an
+        // operator running sprints against more than one Azure DevOps project
+        // may already have that name committed to an unrelated project's PAT.
+        // Without a way to override it, the engine silently provisions THAT
+        // PAT onto the member for this sprint too, clobbering a working
+        // credential and failing every git operation with an auth error that
+        // gives no hint the wrong secret was ever provisioned. Optional --
+        // omitted, provisioning falls back to the provider's own default
+        // exactly as before this flag existed.
+        'azdevops-pat-secret-name': { type: 'string' },
```
kumaakh: **Risk (pre-existing on `main`, extended by this PR):** `azdevopsPatSecretName` already existed in `vcs-auth.mjs` before this PR; this PR extends it with this new public CLI flag plus new call sites in `abort.mjs`, `publish-pr.mjs`, `spawner.mjs`, and `src/supervisor/api.mjs` (this JSON body field), deepening the same violation (and 8 other files: `abort.mjs`, `publish-pr.mjs`, `runner.js`, `bin/cli.mjs`, `spawner.mjs`, `sprint-args.mjs`, `src/supervisor/api.mjs`, plus tests) even though vcs-auth.mjs's own header says runner.js/vcs-auth.mjs "never names or branches on a provider." This flag is the new public entry point PR 497 adds for the whole chain. The need is genuine (can't hardcode a secret name), but the name bakes in today's one PAT provider.

Tracing it: this function already forwards it as an opaque generic `secretName` into `buildProvisionArgsForProvider` -> `impl.buildProvisionArgs({..., secretName})`, a per-provider hook only `vcs-providers/azure-devops.mjs` actually reads. The provider branding exists ONLY at the public boundary (CLI flag/JSON field/param names) -- the plumbing is already generic.

**Suggested change (two parts):**
1. Rename end-to-end: `azdevopsPatSecretName` -> `vcsPatSecretName`, `--azdevops-pat-secret-name` -> `--vcs-pat-secret-name`. Zero internal logic change.
2. For the real generic need (letting a workflow say which stored secret satisfies which of its credential requirements): add an optional provider hook `credentialNeeds()` (e.g. `['pat']` for azure-devops, `[]` for github), and generalize the single string into a map `credentialSecretNames: {pat: 'name'}` via a repeatable `--credential-secret <need>=<name>` flag, forwarded to `buildProvisionArgs` exactly as `secretName` is today.

**Options considered:** (a) rename only -- smallest diff, but a single string doesn't generalize to a provider needing two differently-named secrets later; (b) full credential-needs map -- matches the pattern this codebase already used to fix the identical `DEFAULT_AUTH_MODES` provider-keyed-table problem in `vcs-module.mjs` (`apra-fleet-647.1.5` line of work) and the VCSModule ADR; (c) move secret selection into the credential store's own data model (a `role`/`tag` per secret) -- most automatic long-term, but a larger cross-cutting change than this PR's scope.

**Why (b):** solves the stated generic need using a mechanism already validated in this codebase for the same problem shape, touching the same files this PR already touches -- no larger a diff, just correctly shaped. (a) is an acceptable stopgap only if (b) is filed as an immediate follow-up, not left as debt.

packages/apra-fleet-bridge/src/verbs/preflight.mjs
```diff
@@ -0,0 +1,559 @@
...
+
+  // 1. supervisor-health -- already fetched above (that fetch is the one
+  // throw-worthy step); evaluate what it reported.
+  record('supervisor-health', 'fail', evaluateSupervisorHealth(health));
+
+  // 2. member-free -- GET /api/members. A failure fetching the member list
+  // is reported as this check's own failure, never thrown: it does not make
+  // supervisor-health, member-vcs-provider, credentials, repo-and-base,
+  // pr-capability, beads-health, or playbooks meaningless.
+  const { members, error: membersError } = await fetchMembers(supervisorClient);
```
kumaakh, line 523: **Risk:** 6 independent checks (member fetch, VCS provider, credential, repo/base, beads-health dolt-pull, playbook) run sequentially via separate `await`s despite touching disjoint backends with no shared state, even though this file's own header states the goal is running every check without short-circuiting. Latency is the SUM of each check's I/O time instead of the MAX.

**Suggested change:** Run the independent checks concurrently via `Promise.allSettled`, then assemble the existing report from the settled results in the same order/shape.

**Options considered:** (a) `Promise.all` -- a single throw aborts the rest, losing their results and defeating the "never short-circuit" goal; (b) `Promise.allSettled` over all 6; (c) partial grouping -- unnecessary, no shared state exists between any of the 6.

**Why (b):** direct concurrency-safe equivalent of this file's own stated "run every check, don't stop early" intent, cutting wall-clock time to roughly the slowest single check.

